PR #4, opened 2018-09-14
drop unused dependencies

jschleicher:
Drop the packages we currently do not use from package.xml and headers files in pilz_trajectory_generation.


agutenkunst (changes requested):
Please update the CHANGELOG

jschleicher:
done. Please merge.

agutenkunst (approved):

PR #10, opened 2018-11-12
Hard set last trajectory point to vel and acc zero

agutenkunst:


jschleicher (dismissed):
What about lin? Could this be a helper function in the base class?

Please also add some rationale in the commit message.

agutenkunst:
1. For Lin didn't add it because it was already there :flushed:
2. More generalization: I considered putting it in trajectory_functions.cpp however since the sampling itself is not directly "wrong" I decided to keep it inside the generator to be more expressive.

PR #20, opened 2018-11-19
Fix/zero radius

martiniil:


pilz_trajectory_generation/test/integrationtest_command_list_manager.cpp
```diff
@@ -134,13 +134,14 @@ INSTANTIATE_TEST_CASE_P(InstantiationName, IntegrationTestCommandListManager, ::
...
  *  - Test Sequence:
- *    1. Generate request with two trajectories and zero blend radius.
+ *    1. Generate request with three trajectories and zero blend radius.
  *    2. Generate request with first trajectory and zero blend radius.
  *    3. Generate request with second trajectory and zero blend radius.
  *    4. Generate request with third trajectory and zero blend radius.
  *
  *  - Expected Results:
  *    1. blending is successful, result trajectory is not empty
+ *       TODO: all time steps of resulting trajectory are strictly positive
```
hslusarek: Remove `TODO` statement. Code already implemented.

pilz_trajectory_generation/src/robot_trajectory_helper.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+#include <pilz_trajectory_generation/robot_trajectory_helper.h>
+#include <pilz_trajectory_generation/trajectory_functions.h>
+#include <ros/console.h>
+
+namespace pilz_trajectory_generation
+{
+
+/**
+ * @brief Appends a trajectory to another trajectory leaving out the first state, if it matches the last point of the
```
hslusarek: Move function documentation to header.

pilz_trajectory_generation/src/robot_trajectory_helper.cpp
```diff
@@ -0,0 +1,50 @@
...
+  size_t start_index = 0;
+  if ( (robot_trajectory->getWayPointCount() > 0) && pilz::isRobotStateEqual(robot_trajectory->getLastWayPointPtr(),
+                                                                             tail.getFirstWayPointPtr(),
+                                                                             robot_trajectory->getGroupName(),
+                                                                             ROBOT_STATE_EQUALITY_EPSILON) )
+  {
+    ++start_index;
+  }
+
+  for (size_t i = start_index; i < tail.getWayPointCount(); ++i)
```
hslusarek: In case the last point of the original trajectory and the first point of the tail-trajectory are NOT equal, I suggest using  `robot_trajectory->append(*tail, 0.0);`.

hslusarek: I also suggest that we open an PullRequest on [robot_trajectory](https://github.com/ros-planning/moveit/blob/melodic-devel/moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h) adding an append-function which allows to specify which part of the given trajectory should be appended.

pilz_trajectory_generation/include/pilz_trajectory_generation/robot_trajectory_helper.h
```diff
@@ -0,0 +1,38 @@
...
+namespace pilz_trajectory_generation
+{
+
+class RobotTrajectoryHelper
+{
+public:
+    static void appendWithoutRedundantFirstState(robot_trajectory::RobotTrajectoryPtr &robot_trajectory,
+                                                 robot_trajectory::RobotTrajectory &tail);
+
+    // Constant to check for equality of values.
```
hslusarek: Instead of a comment, I suggest a proper doxygen documentation. It is a public constant after all.

pilz_trajectory_generation/src/robot_trajectory_helper.cpp
```diff
@@ -0,0 +1,50 @@
...
+ * first trajectory.
+ *
+ * @note Controllers, so far at least the ros_controllers::JointTrajectoryController require a timewise strictly
+ * increasing trajectory. If through appending the last point of the original trajectory gets repeated it is removed here.
+ */
+void RobotTrajectoryHelper::appendWithoutRedundantFirstState(robot_trajectory::RobotTrajectoryPtr &robot_trajectory,
+                                                             robot_trajectory::RobotTrajectory &tail)
+{
+  size_t start_index = 0;
+  if ( (robot_trajectory->getWayPointCount() > 0) && pilz::isRobotStateEqual(robot_trajectory->getLastWayPointPtr(),
```
hslusarek: I suggest using `!robot_trajectory->empty()` instead of `robot_trajectory->getWayPointCount() > 0`.

pilz_trajectory_generation/include/pilz_trajectory_generation/robot_trajectory_helper.h
```diff
@@ -0,0 +1,38 @@
...
+
+#include <moveit/robot_trajectory/robot_trajectory.h>
+
+namespace pilz_trajectory_generation
+{
+
+class RobotTrajectoryHelper
+{
+public:
+    static void appendWithoutRedundantFirstState(robot_trajectory::RobotTrajectoryPtr &robot_trajectory,
```
hslusarek: Due to the fact that the function does not manipulate the lifetime of both given trajectories, I suggest to use raw pointers. This is also suggest by the CppCoreGuidelines, see [here](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#r30-take-smart-pointers-as-parameters-only-to-explicitly-express-lifetime-semantics).

pilz_trajectory_generation/include/pilz_trajectory_generation/robot_trajectory_helper.h
```diff
@@ -0,0 +1,38 @@
...
+
+#ifndef ROBOT_TRAJECTORY_HELPER_H
+#define ROBOT_TRAJECTORY_HELPER_H
+
+#include <moveit/robot_trajectory/robot_trajectory.h>
+
+namespace pilz_trajectory_generation
+{
+
+class RobotTrajectoryHelper
```
hslusarek: Instead of introducing another helper class, I suggest working towards an upcoming refactoring. Like I see it, we have three different types of software components in the pilz_trajectory_generation package:
* Firstly, we have the "pure" trajectory generators which take a `MotionPlanRequest` and generate a "simple" linear, cirucular or ptp trajectory.
* Secondly, we have software components which I suggest calling `TrajectoryMerger`. The `TrajectoryBlender` class and your newly created `RobotTrajectoryHelper` belong to this type of software component. Simply put: They take two trajectories and merge them together (one way or another).
* Thirdly, we have the software components which basically use the trajectory generators and trajectory mergers to actually do something. An example for these software components is the `CommandListManager`.
Based on these considerations, I suggest to introduce a base class called `TrajectoryMerger` and derive a specific TrajectoryMerger which basically contains the functionality which you put into your `RobotTrajectoryHelper` class.

hslusarek (changes requested):

pilz_trajectory_generation/include/pilz_trajectory_generation/robot_trajectory_helper.h
martiniil: done.

pilz_trajectory_generation/src/robot_trajectory_helper.cpp
martiniil: done.

pilz_trajectory_generation/include/pilz_trajectory_generation/robot_trajectory_helper.h
martiniil: done.

pilz_trajectory_generation/include/pilz_trajectory_generation/robot_trajectory_helper.h
martiniil: I can see your point and followed your suggestion. Can you please review the new commit?

pilz_trajectory_generation/include/pilz_trajectory_generation/trajectory_merger.h
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * @brief Base class for merging two or more trajectories.
+ */
+class TrajectoryMerger
+{
+public:
+    /**
+     * @brief Merge multiple trajectories into a result trajectory.
+     */
+    virtual void merge(const RobotTrajectoryList& traj_list, robot_trajectory::RobotTrajectory &result) = 0;
```
hslusarek: Due to the fact that a list of trajectories for merging is currently not needed or used, I suggest that we stick to the function signature version you used in `trajectory_appender.cpp`.

hslusarek: If needed we can still add this function later on. Nothing is lost if we don't do it right now.

pilz_trajectory_generation/src/trajectory_appender.cpp
```diff
@@ -0,0 +1,52 @@
...
+{
+  const RobotTrajectoryList traj_list = {tail};
+  merge(traj_list, result);
+}
+
+void TrajectoryAppender::merge(const RobotTrajectoryList &traj_list, robot_trajectory::RobotTrajectory &result)
+{
+  for (auto &traj: traj_list)
+  {
+    moveit::core::RobotStatePtr append_traj_first = std::make_shared<moveit::core::RobotState>(traj.getFirstWayPoint());
```
hslusarek: The need for the creation of an shared pointer only arises because the `isRobotStateEqual` method is not correctly implemented. In `isRobotStateEqual()` we do not want to manipulate the lifetime of the RobotStates given. Therefore, I suggest to add another function signature to the function `isRobotStateEqual()` which only takes const references to the RobotState objects. For simplicity I suggest that we also keep the old function signature of `isRobotStateEqual()`. However, we should mark the old function signature as deprecated. When the new function signature for `isRobotStateEqual()` is added we can remove this unnecessary shared_ptr creation. 

Just a small comment: The shared_ptr used here is also not correct in the sense that we also do not want to manipulate the lifetime of the trajectory.

pilz_trajectory_generation/src/trajectory_appender.cpp
```diff
@@ -0,0 +1,52 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <pilz_trajectory_generation/trajectory_appender.h>
+#include <pilz_trajectory_generation/trajectory_functions.h>
+#include <ros/console.h>
```
hslusarek: Do we need this include?

pilz_trajectory_generation/test/unittest_trajectory_appender.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
```
hslusarek, line 17: `#include <vector>` missing.

hslusarek: `#inlude <string>` missing.

pilz_trajectory_generation/test/unittest_trajectory_appender.cpp
```diff
@@ -0,0 +1,178 @@
...
+ *  2. The whole trajectory is appended.
+ */
+TEST_F(TrajectoryAppenderTest, testMergeWithDifferentFirstPoint)
+{
+  using moveit::core::RobotState;
+  using moveit::core::RobotStatePtr;
+  using robot_trajectory::RobotTrajectory;
+
+  // Step 1
+  double duration_from_previous = 0.1;
```
hslusarek: If it is not changing then change it to a `const double`.

pilz_trajectory_generation/test/unittest_trajectory_appender.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+  traj2.addSuffixWayPoint(robot_state1, duration_from_previous);
+  traj2.addSuffixWayPoint(robot_state2, duration_from_previous);
+
+  // Step 2
+  appender_.merge(traj2, traj1);
+
+  ASSERT_EQ(4u, traj1.getWayPointCount());
+
+  ASSERT_TRUE(pilz::isRobotStateEqual(robot_state1, traj1.getWayPointPtr(0), planning_group_, 1e-12));
```
hslusarek: `1e-12` is a magic number. Use test-global constexpr variable.

pilz_trajectory_generation/test/unittest_trajectory_appender.cpp
```diff
@@ -0,0 +1,178 @@
...
+ *  1. -
+ *  2. The whole trajectory is appended.
+ */
+TEST_F(TrajectoryAppenderTest, testMergeWithDifferentFirstPoint)
+{
+  using moveit::core::RobotState;
+  using moveit::core::RobotStatePtr;
+  using robot_trajectory::RobotTrajectory;
+
+  // Step 1
```
hslusarek: For debugging purposes it is sometimes good to output the start of a test-step as log message.

hslusarek (changes requested):

pilz_trajectory_generation/src/robot_trajectory_helper.cpp
martiniil: https://github.com/PilzDE/moveit/commit/7847d00d89fc5e7945664a3acb61fa43394c8e58

pilz_trajectory_generation/src/trajectory_appender.cpp
martiniil: No :)

pilz_trajectory_generation/test/unittest_trajectory_appender.cpp
martiniil: Done.

pilz_trajectory_generation/test/unittest_trajectory_appender.cpp
martiniil: Done.

pilz_trajectory_generation/test/unittest_trajectory_appender.cpp
martiniil: Done.

pilz_trajectory_generation/test/unittest_trajectory_appender.cpp
martiniil: Done.

pilz_trajectory_generation/src/trajectory_appender.cpp
martiniil: See https://github.com/PilzDE/pilz_industrial_motion/pull/21

hslusarek (approved):

PR #24, opened 2018-11-21
Fix/add sequence tests

agutenkunst:


pilz_trajectory_generation/test/integrationtest_command_list_manager.cpp
```diff
@@ -217,6 +212,78 @@ TEST_P(IntegrationTestCommandListManager, concatThreeSegments)
...
 
+/**
+ * @brief Integration test for the concatenation of two ptp commands
+ * Sends a pilz_msgs::MotionSequenceRequest request.
+ *
+ *  - Test Sequence:
+ *    1. Generate request with two PTP trajectories and zero blend radius.
+ *
+ *  - Expected Results:
+ *    1. blending is successful, result trajectory is not empty
```
martiniil: Actually, no blending is performed. Please change this comment in all tests with zero blend radius. E.g. something like:
> Generation of concatenated trajectory is successful

martiniil (changes requested):

martiniil (approved):

PR #29, opened 2018-11-26
Remove error message of helper function

martiniil:
The trajectory function isRobotStateEqual() printed an error, if the result of comparison is false.
This happened with every blend command.


agutenkunst (dismissed):
I would suggest removing all output from such a general function.

martiniil:
With the debug output you can tell, which part of the robot_states is unequal. I find this quite useful. After all the number of calls to this function is very limited.

agutenkunst (approved):

PR #33, opened 2018-11-29
Fix/circ joint goal feature

martiniil:
Circ goal can as well be given as joint state.

### DIFF
--- a/pilz_robot_programming/test/integrationtests/tst_api_cmd_conversion.py
+++ b/pilz_robot_programming/test/integrationtests/tst_api_cmd_conversion.py
@@ -511,15 +511,10 @@ class TestAPICmdConversion(unittest.TestCase):
 
             Test sequence:
                 1. Call circ convert function with center point.
-                2. Call circ convert function with interim point.
 
 
             Test Results:
                 1. Correct MotionPlanRequest is returned.
-                2. Correct MotionPlanRequest is returned.
-                3. None is returned.
-                4. None is returned.
-                5. None is returned.
         """
         exp_goal_pose = self.test_data.get_pose("LINPose1", PLANNING_GROUP_NAME)
         exp_help_pose = self.test_data.get_pose("CIRCCenterPose", PLANNING_GROUP_NAME)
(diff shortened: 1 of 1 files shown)


pilz_robot_programming/test/integrationtests/tst_api_cmd_conversion.py
```diff
@@ -551,6 +546,29 @@ def test_circ_cmd_convert_interim(self):
         self._analyze_request_pose(TARGET_LINK_NAME, exp_goal_pose, req)
         self._analyze_request_circ_help_point("interim", exp_help_pose, req)
 
+    def test_circ_cmd_convert_joint(self):
```
hslusarek, line 549: You test the API. But who tests that the newly created request type actually can be executed?

pilz_robot_programming/src/pilz_robot_programming/commands.py
```diff
@@ -458,7 +458,8 @@ class Circ(_BaseCmd):
         The circle can be completely defined by stating a interim `or` an center position.
         However, only one of both should be stated.
 
-    :param goal: The goal of the motion stated in Cartesian space (geometry_msgs/Pose).
+    :param goal: The goal of the motion, which can be given in joint (list of float, in the order of active joints in
```
hslusarek: What about the center or interim point? Can they also be stated as joint values?

hslusarek: In case it is possible, please add test to check this.

hslusarek (changes requested):

pilz_robot_programming/src/pilz_robot_programming/commands.py
martiniil: I agree that it is inconsistent to introduce the joint state option only for the goal.
Reverting this change ...

pilz_robot_programming/test/integrationtests/tst_api_cmd_conversion.py
martiniil: Not relevant any more, since we do not officially include the feature.

PR #36, opened 2018-12-04
add pilz_robot_programming to meta-package-depends

jschleicher:
to install it together with `apt install pilz-industrial-motion`.
Changelogs updated.


martiniil (approved):
Thanks for the changelog update.

PR #37, opened 2018-12-04
Fix/ik self collision/issue 29

rfeistenauer:
Fixes the problem, that self collision only throws error instead of searching for an alternative solution.


pilz_trajectory_generation/test/unittest_trajectory_functions.cpp
```diff
@@ -134,17 +135,32 @@ bool TrajectoryFunctionsTest::tfNear(const Eigen::Affine3d& pose1, const Eigen::
...
+class TrajectoryFunctionsTestOnlyGripper: public TrajectoryFunctionsTestBase {};
+
 // Instantiate the test cases for robot model with and without gripper
-INSTANTIATE_TEST_CASE_P(InstantiationName, TrajectoryFunctionsTest, ::testing::Values(
+INSTANTIATE_TEST_CASE_P(InstantiationName, TrajectoryFunctionsTestFlangeAndGripper, ::testing::Values(
                           PARAM_MODEL_NO_GRIPPER_NAME,
                           PARAM_MODEL_WITH_GRIPPER_NAME
                           ));
 
+// Instantiate the test cases for robot model with and without gripper
```
jschleicher: comment: copy'n'paste error?

pilz_trajectory_generation/test/unittest_trajectory_functions.cpp
```diff
@@ -25,6 +25,7 @@
 #include <moveit/robot_model_loader/robot_model_loader.h>
 #include <moveit/robot_model/robot_model.h>
 #include <moveit/robot_model/joint_model_group.h>
+#include <moveit/planning_scene/planning_scene.h>
```
jschleicher, line 28: Where is the include `planning_scene` used?

pilz_trajectory_generation/include/pilz_trajectory_generation/trajectory_functions.h
```diff
@@ -54,7 +55,8 @@ bool computePoseIK(const robot_model::RobotModelConstPtr& robot_model,
                    const std::map<std::string, double>& seed,
                    std::map<std::string, double>& solution,
                    bool check_self_collision = true,
-                   int max_attempt = 2);
+                   int max_attempt = 10,
+                   double timeout = 0.005);
```
jschleicher: is this in seconds? Are 5 milliseconds a good default for all platforms?
Usually named constants are preferred over magic numbers.

pilz_trajectory_generation/include/pilz_trajectory_generation/trajectory_functions.h
```diff
@@ -221,6 +224,23 @@ bool intersectionFound(const Eigen::Vector3d &p_center,
                        const Eigen::Vector3d &p_current,
                        const Eigen::Vector3d &p_next,
                        const double& r);
+
+/**
+ * @brief Checks if current robot state is in self collision.
+ * @param test_for_self_collision Flag to deactivate this check during IK.
```
jschleicher, line 229: documentation does not match the function signature.

pilz_trajectory_generation/src/trajectory_functions.cpp
```diff
@@ -54,31 +53,18 @@ bool pilz::computePoseIK(const moveit::core::RobotModelConstPtr &robot_model,
   // set the seed
   rstate.setVariablePositions(seed);
 
+  moveit::core::GroupStateValidityCallbackFn ik_constraint_function;
+  ik_constraint_function = boost::bind(&pilz::isStateColliding, check_self_collision, robot_model, _1, _2, _3);
+
   // call ik
   // TODO: Should consider self collision already.
```
jschleicher: could you remove the TODO note now?

pilz_trajectory_generation/test/unittest_trajectory_functions.cpp
```diff
@@ -417,6 +426,107 @@ TEST_P(TrajectoryFunctionsTest, testComputePoseIKInvalidFrameId)
...
+  std::map<std::string, double> ik_actual1;
+  EXPECT_TRUE(pilz::computePoseIK(robot_model_,
+                                  planning_group_,
+                                  tcp_link_,
+                                  pose_expect,
+                                  frame_id,
+                                  ik_seed,
+                                  ik_actual1,
+                                  false,
+                                  1));
```
jschleicher: Having a closer look this seems to rely on odd behaviour of the IK plugin. There was an upstream [PR](https://github.com/ros-planning/moveit/pull/585) and I think, we should re-generate the ikfast_plugin from the current template there. Then the collision-free solution should be delivered on the first attempt (without involving random seeds making the number of calls non-deterministic).

Could you think of another way to implement this "fail without collision check"-test?

jschleicher (changes requested):
Please address the comments.

Did you check code coverage and static analysis with klocwork or should I do that as part of the review?

pilz_trajectory_generation/include/pilz_trajectory_generation/trajectory_functions.h
rfeistenauer: the 5 ms were suggested somewhere on the web. 10 * 5 ms => min 20 Hz which was supposed to be an acceptable speed. Don't think i can find the quote quickly. The max attempts have been magic number already, thought you preferred it in the header for code api reasons maybe. Can change it if neccessary

rfeistenauer:
Checked code coverage. Its 100%
Never did the klocwork check so i would suggest either you do it, or you point me to a tutorial / explain it.

jschleicher (approved):

PR #38, opened 2018-12-11
Convert Eigen::Affine3d -> Eigen::Isometry3d.

clalancette:
This is to keep up with the recent changes in moveit
(https://github.com/ros-planning/moveit/pull/1096) and
keep this package building in melodic (it is currently failing; see http://build.ros.org/view/Mbin_uB64/job/Mbin_uB64__pilz_industrial_motion_testutils__ubuntu_bionic_amd64__binary/)

### DIFF
--- a/pilz_industrial_motion_testutils/include/pilz_industrial_motion_testutils/motion_plan_request_director.h
+++ b/pilz_industrial_motion_testutils/include/pilz_industrial_motion_testutils/motion_plan_request_director.h
@@ -75,7 +75,7 @@ private:
    * @param pose
    * @return
    */
-  Eigen::Affine3d rawQuatVectorToEigen(const std::vector<double> &pose);
+  Eigen::Isometry3d rawQuatVectorToEigen(const std::vector<double> &pose);
 
   /**
    * @brief Construct a robot state from the start joint values in cmd.
--- a/pilz_industrial_motion_testutils/src/motion_plan_request_director.cpp
+++ b/pilz_industrial_motion_testutils/src/motion_plan_request_director.cpp
@@ -124,7 +124,7 @@ moveit_msgs::MotionPlanRequest MotionPlanRequestDirector::getCIRCCartReq(const m
   return builder.getRequest();
 }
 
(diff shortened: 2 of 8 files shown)


clalancette:
@rhaschke FYI (I think you've been doing most of this work in MoveIt proper)

rhaschke (approved):
MoveIt migrated from Eigen::Affine3d to more efficient Eigen::Isometry3d.
Travis seems to fail due to an unrelated issue.

jschleicher:
Thank you for bringing this up and thanks for your pull request.

The travis failure seems to be due to the [ikfast_plugin package failing to build](http://build.ros.org/job/Mbin_uB64__prbt_ikfast_manipulator_plugin__ubuntu_bionic_amd64__binary/11/console) with the latest moveit upload as well. So we'll have to re-generate the plugin from the upstream template and do a binary release.

rhaschke:
Re-generating the ikfast plugin will not resolve the issue. The ikfast generator is not yet adapted to the new IK plugin API (@mlautman: reminder for https://github.com/ros-planning/moveit/issues/1262). 
I think, for now, it will suffice to accept deprecation warnings and remove "-Werror=deprecated-declarations" in the ikfast build.
If you prefer to cleanup the ikfast API, please consult the [migration notes](https://github.com/ros-planning/moveit/blob/melodic-devel/MIGRATION.md) for hints and [KDL plugin](https://github.com/ros-planning/moveit/tree/melodic-devel/moveit_kinematics/kdl_kinematics_plugin) for an example.

clalancette:
@jschleicher If you don't mind disabling warnings for the time being, that would be ideal.  I'd like to do a Melodic sync soon (next week), but the regression on these packages (via the MoveIt! changes) is holding that back.  Let me know what you want to do, and thanks for your effort.

jschleicher:
@clalancette Thanks for taking care of the release. Yes, we'll try a test build and push a point release of our packages tomorrow morning. So a melodic sync next tuesday (allowing us a quick test of the binaries on monday) should be feasible.

PR #45, opened 2019-01-04
Increase too optimistic IK timeout causing unstable tests in CI

agutenkunst:
Make sure that services are up before using them. Tests need to be aborted if not.

### DIFF
--- a/pilz_trajectory_generation/test/integrationtest_command_planning.cpp
+++ b/pilz_trajectory_generation/test/integrationtest_command_planning.cpp
@@ -135,11 +135,11 @@ TEST_F(IntegrationTestCommandPlanning, PTPJointGoal)
   srv.request.motion_plan_request.group_name = planning_group_;
   srv.request.motion_plan_request.planner_id = "PTP";
 
-  ros::service::waitForService(PLAN_SERVICE_NAME, ros::Duration(10));
+  ASSERT_TRUE(ros::service::waitForService(PLAN_SERVICE_NAME, ros::Duration(testutils::DEFAULT_SERVICE_TIMEOUT)));
   ros::ServiceClient client = node_handle.serviceClient<moveit_msgs::GetMotionPlan>(PLAN_SERVICE_NAME);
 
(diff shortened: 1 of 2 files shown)


pilz_trajectory_generation/test/integrationtest_command_planning.cpp
```diff
@@ -223,7 +223,7 @@ TEST_F(IntegrationTestCommandPlanning, PTPPoseGoal)
   srv.request.motion_plan_request.group_name = planning_group_;
   srv.request.motion_plan_request.planner_id = "PTP";
 
-  ros::service::waitForService(PLAN_SERVICE_NAME, ros::Duration(10));
+  ASSERT_TRUE(ros::service::waitForService(PLAN_SERVICE_NAME, ros::Duration(10)));
```
martiniil: Use testutils::DEFAULT_SERVICE_TIMEOUT everywhere?

martiniil (changes requested):

pilz_trajectory_generation/test/integrationtest_command_planning.cpp
agutenkunst: Done.

martiniil:
In log of https://travis-ci.org/PilzDE/pilz_industrial_motion/jobs/477346308:

> [DEBUG] [1547041875.423349531]: searchPositionIK2: Position request pose is 0.313627 0.178519 0.355369 -1 -1.16029e-07 5.43542e-07 9.09851e-08
> [DEBUG] [1547041875.423390566]: IK timed out

In this case a second attempt uses random seed values, leading to our problem:
http://docs.ros.org/kinetic/api/moveit_core/html/robot__state_8cpp_source.html#l01619

So far my understanding. The timeout is defined in pilz_trajectory_generation/trajectory_functions.h:
`const double timeout = 0.005`

martiniil (approved):

martiniil:
@agutenkunst Do you approve these changes, too?

PR #47, opened 2019-01-07
Enhance documentation on sequence

agutenkunst:
* Move documentation on conditions for a sequence request to be
  valid into the public interface


pilz_trajectory_generation/include/pilz_trajectory_generation/command_list_manager.h
```diff
@@ -55,11 +60,8 @@ class CommandListManager {
 
 private:
   /**
-   * @brief Validate if the request list fullfills the following conditions
-   * - All request are about the same group
-   * - All blending radii are non negative
-   * - The blending radius of the last request is 0
-   * - Only the first request has a start state
+   * @brief Validate if the request list fullfills the following conditions noted
```
martiniil: I think we should omit "following" here.

martiniil (changes requested):
Thanks. We can merge this after small correction.

martiniil (approved):

PR #55, opened 2019-01-22
Fix handling of empty sequence in capabilities

agutenkunst:
* Planning + Execution of an empty sequence should be
  defined as successful to avoid errors on clients.
  However a warning indicates to the user that sending
  an empty sequence could have been unintentionally.


martiniil (approved):
Very much approved!

PR #66, opened 2019-01-30
Fix/instance not cleared

martiniil:


pilz_robot_programming/src/pilz_robot_programming/robot.py
```diff
@@ -357,29 +351,35 @@ def _check_version(self, version):
...
-            rospy.logerr("An instance of Robot class already exists.")
-            self._ctor_exception_flag = True
-            raise RobotMultiInstancesError("Only one instance of Robot class can be created!")
-        else:
-            rospy.set_param(self._SINGLE_INSTANCE_FLAG, True)
+        if rospy.has_param(self._INSTANCE_PID):
+            pid = rospy.get_param(self._INSTANCE_PID)
+            # check for the existence of pid
+            try:
+                os.kill(pid, 0)
```
agutenkunst: What signal does 0 correspond to ?

pilz_robot_programming/src/pilz_robot_programming/robot.py
```diff
@@ -360,26 +364,33 @@ def _check_version(self, version):
             raise RobotVersionError("Version of Robot API does not match! "
                                     "Current installed version is " + __version__ + "!")
 
+    def _claim_single_instance(self):
+        # check if an instance exists
+        self._check_single_instance()
+
+        # If no exception has been raised, the pid and create_time is stored (overwrites old one)
```
agutenkunst: if(_check_single_instance()) would be more readable, then throw on false...

agutenkunst (changes requested):

pilz_robot_programming/src/pilz_robot_programming/robot.py
martiniil: It has no symbolic name but is used to check if a process is running.
https://wiki.ubuntuusers.de/Signale/

martiniil:
I would like to squash some commits when you approved it.

agutenkunst (approved):

PR #74, opened 2019-01-31
Update blending algorithm documentation

agutenkunst:
* Use article as class
* Add company name as author
* Add date on every page
* Remove old template artefacts


martiniil (approved):
Thank you

PR #80, opened 2019-02-04
Add missing forwarding of target_link in _get_goal_pose

agutenkunst:


pilz_robot_programming/src/pilz_robot_programming/commands.py
```diff
@@ -226,7 +227,9 @@ def _get_sequence_request(self, robot):
 
     def _get_goal_pose(self, robot):
         """Determines the goal pose for the given command."""
-        current_pose = robot.get_current_pose(base=self._reference_frame if self._reference_frame else "prbt_base")
+        current_pose = robot.get_current_pose(
+            target_link=self._target_link if self._target_link else _DEFAULT_TARGET_LINK,
```
martiniil: If we really need `if self._target_link`, we could already insert it in the `__init__` function. `_DEFAULT_TARGET_LINK` is already the default there. Else I would have to do `if self._target_link` everytime I use it and I would not be happy with that.

martiniil: `_DEFAULT_BASE_LINK` should be reused in robot.py. But strictly speaking this is not subject of this PR.

martiniil (changes requested):

martiniil (approved):

PR #87, opened 2019-02-13
Fix/circ path generation

martiniil:
The formula in pilz_trajectory_generation/src/path_circle_generator.cpp:

> const KDL::Vector center_point = start_pose.p + (u*dot(t,t)*dot(u,v) - t*dot(u,u)*dot(t,v))*0.5/pow(w.Norm(),2);

is only valid if the plane defined by start, interim and goal is well-defined. The vector `w` represents a normal on this plane and should be greater than some epsilon.


pilz_trajectory_generation/src/path_circle_generator.cpp
```diff
@@ -71,9 +71,13 @@ std::unique_ptr<KDL::Path> PathCircleGenerator::circleFromInterim(
...
+  const KDL::Vector w = t*u; // cross product
 
   // circle center
+  if (w.Norm() < MAX_COLINEAR_NORM)
+  {
+    throw KDL::Error_MotionPlanning_Circle_No_Plane();
+  }
   const KDL::Vector center_point = start_pose.p + (u*dot(t,t)*dot(u,v) - t*dot(u,u)*dot(t,v))* 0.5/pow(w.Norm(),2);
 
   // compute the rotation angle
```
agutenkunst, line 83: Please take a look at the comment here and the ones in the following lines. Maybe put some things here in subfunctions...

agutenkunst (changes requested):

agutenkunst (approved):

Issue #99, opened 2019-02-22
Potential memory leak in path_circle_generator


agutenkunst:
https://github.com/PilzDE/pilz_industrial_motion/blob/1657377f87f524401c6b7d2228bd39e9a037043d/pilz_trajectory_generation/src/path_circle_generator.cpp#L42-L58

and

https://github.com/PilzDE/pilz_industrial_motion/blob/1657377f87f524401c6b7d2228bd39e9a037043d/pilz_trajectory_generation/src/path_circle_generator.cpp#L94-L110

look like they could leak...

martiniil:
https://help.github.com/en/articles/closing-issues-using-keywords

PR #100, opened 2019-02-22
Related to #99. Clarifies that there is no leak.

agutenkunst:


pilz_trajectory_generation/src/path_circle_generator.cpp
```diff
@@ -41,15 +41,14 @@ std::unique_ptr<KDL::Path> PathCircleGenerator::circleFromCenter(
   KDL::RotationalInterpolation* rot_interpo = new KDL::RotationalInterpolation_SingleAxis();
   try
   {
-    KDL::Path* circ = new KDL::Path_Circle(start_pose,
+    return std::unique_ptr<KDL::Path>(new KDL::Path_Circle(start_pose,
```
gavanderhoorn, line 44: `std::make_unique<..>(..)`?

agutenkunst: c++11 ...

gavanderhoorn: https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique

https://stackoverflow.com/questions/7038357/make-unique-and-perfect-forwarding

Could bring/roll your own.

martiniil: Ah, I didn't know that. Thanks for the hint.

martiniil (approved):
The value of implementing a surrogate for a missing STL function is debatable. Still merging ...

PR #106, opened 2019-02-28
Ensure valid circ aux point if possible

martiniil:
KDL::Path_Circle computes a plane from start, center and auxiliary point.
So the interim point is not suited, if start, center and interim are colinear.

- [x] check that using goal as aux point is valid
- [x] implement test reproducing this issue

### DIFF
--- a/pilz_trajectory_generation/CHANGELOG.rst
+++ b/pilz_trajectory_generation/CHANGELOG.rst
@@ -5,6 +5,8 @@ Changelog for package pilz_trajectory_generation
 Forthcoming
 -----------
 * update dependencies of trajectory_generation
+* fix CIRC path generator and increase test coverage
+* Contributors: Pilz GmbH and Co. KG
 
 0.3.6 (2019-02-26)
 ------------------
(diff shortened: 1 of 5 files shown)


hslusarek:
@agutenkunst Please have a look at this merge request. You also done a lot of work with regard to the circ command. Please check if you have done something similar... Please be also aware of the fact that this merge request, will merge into kinetic and not melodic.

ct2034 (changes requested):
Command is executed succefully but ends up at wrong point. See https://github.com/ct2034/my_pilz_sandbox/blob/f69e6fa81c501d852623f4a3215bef6150ad5261/my_pilz_sandbox/scripts/circ.py

martiniil:
Actually, there was a bug in testutils (see #125) preventing the circ test from failing.

ct2034 (approved):

ct2034:
succesfully tested with https://github.com/ct2034/my_pilz_sandbox/blob/f481afe227351fa657c69c789c8149cbaa83cf50/my_pilz_sandbox/scripts/circ.py

PR #108, opened 2019-03-01
Make sure to specify the maximum number of attempts.

clalancette:
This makes it equivalent to the default, which is 0.

This should fix the current regression on the buildfarm: http://build.ros.org/job/Mbin_uB64__pilz_trajectory_generation__ubuntu_bionic_amd64__binary/44/consoleFull

Note that this *may* need to change again with the moveit 1.0 release, but we'll cross that bridge after the current sync.

It would be nice to merge this and then get a new release into Melodic ASAP so I can do a sync next week.  Thanks in advance.


martiniil:
I see ... I hope we will be fine, if my colleagues tackle this on monday???
fyi @agutenkunst @hslusarek

martiniil:
This should work: https://github.com/ros/rosdistro/pull/20418

clalancette:
> This should work: [ros/rosdistro#20418](https://github.com/ros/rosdistro/pull/20418)

Thanks for the quick turn around.

Issue #115, opened 2019-03-14
Using super in pilz robot programming


rfeistenauer:
To allow multiple inheritance and dependency injection (e.g. to mock in tests) one should use super to call parents init instead of directly calling it. Additionally **kwargs has to be accepted, otherwise it will throw issues, when using multiple inheritance.

PR #116, opened 2019-03-14
using super in robot and commands

rfeistenauer:
This should fix Issue #115

### DIFF
--- a/pilz_robot_programming/src/pilz_robot_programming/robot.py
+++ b/pilz_robot_programming/src/pilz_robot_programming/robot.py
@@ -105,7 +105,8 @@ class Robot(object):
     _PID_STRING = "pid"
     _PROCESS_CREATE_TIME_STRING = "create_time"
 
-    def __init__(self, version=None):
+    def __init__(self, version=None, **kwargs):
+        super(Robot, self).__init__()
         rospy.logdebug("Initialize Robot Api.")
 
         # tf listener is necessary for pose transformation
--- a/pilz_robot_programming/src/pilz_robot_programming/robot.py
+++ b/pilz_robot_programming/src/pilz_robot_programming/robot.py
@@ -105,8 +105,8 @@ class Robot(object):
     _PID_STRING = "pid"
     _PROCESS_CREATE_TIME_STRING = "create_time"
 
-    def __init__(self, version=None, **kwargs):
-        super(Robot, self).__init__()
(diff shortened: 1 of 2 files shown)


pilz_robot_programming/src/pilz_robot_programming/robot.py
```diff
@@ -105,7 +105,8 @@ class Robot(object):
     _PID_STRING = "pid"
     _PROCESS_CREATE_TIME_STRING = "create_time"
 
-    def __init__(self, version=None):
+    def __init__(self, version=None, **kwargs):
+        super(Robot, self).__init__()
```
rfeistenauer: probalby it should even be 
```
def __init__(version=None, *args, **kwargs)
    super(Robot, self).__init__(*args, **kwargs)
```

agutenkunst:
There are checks failing. Could you check on this?

agutenkunst:
There is a typo in the first commit and maybe add "Closes" for automatic issue closing in your commit.

rfeistenauer:
```
class Ptp(_BaseCmd):
    """
    ...
    """
    def __init__(self, acc_scale=None, vel_scale=None, *args, **kwargs):      
        acc_scale_final = acc_scale if acc_scale is not None else Ptp._calc_acc_scale(vel_scale)        
        super(Ptp, self).__init__(acc_scale=acc_scale_final, vel_scale=vel_scale, *args, **kwargs)
```
Depending on whether you want it for documentation reasons or not, you could even throw out all kwargs out of the command classes Ptp, Lin and Circ. They will still show on code completion hint since super takes care of that, but the doc above the classes may look off place.

ct2034 (changes requested):
I think it would make sense to also remove the explicit params in the inheriting classes (when calling inti with super)

ct2034 (approved):

ct2034:
@dbakovic @agutenkunst @hslusarek @martiniil 
Do we want this? I think it is a lot more clean

martiniil (commented):
I don't see why someone would inherit from the Robot class multiple times. For single inheritance super() in the base class has no effect, right?
 
@rfeistenauer What are your thoughts on that?

martiniil:
I got it.

rfeistenauer:
just for documentation purpose: https://www.youtube.com/watch?v=EiOglTERPEo answeres this question

PR #112, opened 2019-03-13
update dependencies of trajectory_generation

jschleicher:
* convert package.xml to format 2 (See http://docs.ros.org/jade/api/catkin/html/howto/format2/migrating_from_format_1.html)
* make eigen_conversions a `<depend>` to install it at runtime (Closes #110)
* drop prbt_ikfast_manipulator_plugin; trajectory_generation runtime packages
  are independent of the robot used and the corresponding moveit_config is
  already listed as test_depend.
* degrade pilz_testutils to `<test_depend>`, since in format 2 this also
  fulfils the dependency to build the test (Reverts 421c8af)
* moveit_commander isn't used in trajectory_generation, move it into
  the _testutils package
* drop message_runtime dependency, sind moveit_msgs etc depend on it
  and install it implicitly

PR #117, opened 2019-03-14
update dependencies of trajectory_generation

### LINKED ISSUE #110: melodic-ros-base installation: Could not load library...libeigen_conversions.so
### Commit
ros-melodic-pilz-trajectory-generation       0.4.1-1bionic.20190302.180259      armhf

### Steps to reproduce
How can the issue/bug be reproduced?
1. sudo apt install ros-melodic-ros-base 
2. roslaunch prbt_moveit_config moveit_planning_execution.launch pipeline:=pilz_command_planner

### Observed behavior
Error string: Could not load library (Poco exception = libeigen_conversions.so: cannot open shared object file: No such file or directory

### Solution
sudo apt install libtf-conversions-dev

### Reference to test
If there is no test that can detect the issue/bug, please create one and
reference it here.

jschleicher:
* add eigen_conversions as `<run_depend>` to install it at runtime (Closes #110)


martiniil:
I prefer to merge #112.

PR #111, opened 2019-03-13
Fix/adapt tests to new testdata

agutenkunst:

PR #118, opened 2019-03-14
Fix/circ larger 180deg test

agutenkunst:
To be merge after https://github.com/PilzDE/pilz_industrial_motion/pull/111 !

Closes gap in coverage where https://github.com/PilzDE/pilz_industrial_motion/blob/685c781874d1771b67058b040983813d9185a719/pilz_trajectory_generation/src/path_circle_generator.cpp#L93 was not covered


martiniil:
No longer needed due to #106

PR #121, opened 2019-03-18
using getSingleDOFJointModels instead of  getActiveJointModels to support mimic joints

mmoerdijk:
Using getSingleDOFJointModels instead of  getActiveJointModels to support mimic joints. Without this change the planner crashes when looking for the joint limits of the mimic joints.


jschleicher:
What configuration are you using? Could you add the config where the planner crashes?
Best would be a test case that fails with the current implementation and passes with your fix.

mmoerdijk:
Good idea I will look into making a simple test case

jschleicher:
Thank you!

ct2034 (changes requested):
@mmoerdijk 

@jschleicher:
> What configuration are you using? Could you add the config where the planner crashes?
> Best would be a test case that fails with the current implementation and passes with your fix.

agutenkunst:
@mmoerdijk ping

martiniil:
No progress since 2 months. Closing for now. Feel free to re-open.

PR #129, opened 2019-03-26
Remove genmsg from 'pilz_msgs' package: it's not needed

gavanderhoorn:
The `genmsg` package is not needed as a dependency of `*_msgs` packages. `message_runtime` and `message_generation` are responsible for pulling in the necessary dependencies.


gavanderhoorn:
As I did this using the GH web UI, there are two commits. They can (and should) be squashed if/when this PR is accepted.

hslusarek (approved):

PR #131, opened 2019-03-29
Fix/Refactor: TrajectoryGeneratorPTP, JointLimitsContainer

martiniil:
* Fix handling of non-equal joint limits in PTP planner
* Adjust tests affected by the changes (needs [pilz_robots#94](https://github.com/PilzDE/pilz_robots/pull/94), therefore no CI before shadow release)
* Re-establish 100% line coverage of pilz_trajectory_generation

### DIFF
--- a/pilz_trajectory_generation/include/pilz_trajectory_generation/joint_limits_container.h
+++ b/pilz_trajectory_generation/include/pilz_trajectory_generation/joint_limits_container.h
@@ -133,8 +133,8 @@ private:
    * @param joint_limit
    * @param common_limit the current most strict limit
    */
-  void updateCommonLimit(const pilz_extensions::JointLimit& joint_limit,
-                         pilz_extensions::JointLimit& common_limit) const;
+  static void updateCommonLimit(const pilz_extensions::JointLimit& joint_limit,
+                                pilz_extensions::JointLimit& common_limit);
 
 protected:
   /// Actual container object containing the data
--- a/pilz_trajectory_generation/include/pilz_trajectory_generation/joint_limits_container.h
+++ b/pilz_trajectory_generation/include/pilz_trajectory_generation/joint_limits_container.h
@@ -60,18 +60,30 @@ public:
   bool empty() const;
 
   /**
-   * @brief Returns joint limit fusion of all(position, velocity, acceleration, deceleration) limits
-   * There are cases where the common limits of all is needed. This function calculate and return a joint_limit
-   * which respects all limits of the container.
-   * @return fused joint_limit
-   * if there are no limits, the flag has_[position|velocity|...]_limits is set to false.
+   * @brief Returns joint limit fusion of all(position, velocity, acceleration, deceleration) limits for all joint.
+   * There are cases where the most strict limit of all limits is needed.
+   * If there are no matching limits, the flag has_[position|velocity|...]_limits is set to false.
+   *
+   * @return joint limit
    */
   pilz_extensions::JointLimit getCommonLimit() const;
 
   /**
-   * @brief getLimit get the limit with the name,
+   * @brief Returns joint limit fusion of all(position, velocity, acceleration, deceleration) limits for given joints.
+   * There are cases where the most strict limit of all limits is needed.
+   * If there are no matching limits, the flag has_[position|velocity|...]_limits is set to false.
+   *
+   * @param joint_names
+   * @return joint limit
+   * @throws std::out_of_range if a joint limit with this name does not exist
+   */
+  pilz_extensions::JointLimit getCommonLimit(const std::vector<std::string> &joint_names) const;
+
+  /**
+   * @brief getLimit get the limit for the given joint name
    * @param joint_name
-   * @return joint_limit, throws std::out_of_range if a joint limit with this name does not exist
+   * @return joint limit
+   * @throws std::out_of_range if a joint limit with this name does not exist
    */
   pilz_extensions::JointLimit getLimit(const std::string& joint_name) const;
 
@@ -114,6 +126,15 @@ public:
    */
   bool verifyPositionLimits(const std::vector<std::string> &joint_names,
                             const std::vector<double> &joint_positions) const;
+
+private:
+  /**
+   * @brief update the most strict limit with given joint limit
+   * @param joint_limit
+   * @param common_limit the current most strict limit
+   */
+  void updateCommonLimit(const pilz_extensions::JointLimit& joint_limit,
+                         pilz_extensions::JointLimit& common_limit) const;
 
 protected:
   /// Actual container object containing the data
--- a/pilz_trajectory_generation/src/joint_limits_container.cpp
+++ b/pilz_trajectory_generation/src/joint_limits_container.cpp
@@ -20,141 +20,75 @@
 #include "ros/ros.h"
 #include <stdexcept>
 
-bool pilz::JointLimitsContainer::addLimit(const std::string &joint_name, pilz_extensions::JointLimit joint_limit)
+namespace pilz
+{
+
+bool JointLimitsContainer::addLimit(const std::string &joint_name, pilz_extensions::JointLimit joint_limit)
 {
   if(joint_limit.has_deceleration_limits && joint_limit.max_deceleration >= 0)
   {
     ROS_ERROR_STREAM("joint_limit.max_deceleration MUST be negative!");
     return false;
   }
-  else
+  if (!container_.insert(std::pair<std::string, pilz_extensions::JointLimit>(joint_name, joint_limit)).second)
   {
-    container_.insert(std::pair<std::string, pilz_extensions::JointLimit>(joint_name, joint_limit));
-    return true;
+    ROS_ERROR_STREAM("joint_limit for joint " << joint_name << " already contained.");
+    return false;
   }
-
+  return true;
 }
 
-bool pilz::JointLimitsContainer::hasLimit(const std::string &joint_name) const
+bool JointLimitsContainer::hasLimit(const std::string &joint_name) const
 {
   return container_.find(joint_name) != container_.end();
 }
 
-size_t pilz::JointLimitsContainer::getCount() const
+size_t JointLimitsContainer::getCount() const
 {
   return container_.size();
 }
 
-bool pilz::JointLimitsContainer::empty() const
+bool JointLimitsContainer::empty() const
 {
   return container_.empty();
 }
 
-pilz_extensions::JointLimit pilz::JointLimitsContainer::getCommonLimit() const
+pilz_extensions::JointLimit JointLimitsContainer::getCommonLimit() const
 {
   pilz_extensions::JointLimit common_limit;
-
-  if(container_.empty())
+  for(auto it = container_.begin(); it != container_.end(); ++it)
   {
-    common_limit.has_position_limits = false;
-    common_limit.min_position = 0;
-    common_limit.max_position = 0;
-    common_limit.has_velocity_limits = false;
-    common_limit.max_velocity = 0;
-    common_limit.has_acceleration_limits = false;
-    common_limit.max_acceleration = 0;
-    common_limit.has_deceleration_limits = false;
-    common_limit.max_deceleration = 0;
-    return common_limit;
+    updateCommonLimit(it->second, common_limit);
   }
-
-  common_limit = container_.begin()->second;
-
-  for(auto it = std::next(container_.begin()); it != container_.end(); ++it)
-  {
-
-    // If this limit has position limits
-    if(it->second.has_position_limits)
-    {
-      // Merge if common_limit has allready limit
-      if(common_limit.has_position_limits)
-      {
-        common_limit.min_position = std::max(common_limit.min_position, it->second.min_position);
-        common_limit.max_position = std::min(common_limit.max_position, it->second.max_position);
-      }
-      else
-      {
-        common_limit.has_position_limits = true;
-        common_limit.min_position = it->second.min_position;
-        common_limit.max_position = it->second.max_position;
-      }
-    }
-
-    // If this limit has velocity limits
-    if(it->second.has_velocity_limits)
-    {
-      // Merge if common_limit has allready limit
-      if(common_limit.has_velocity_limits)
-      {
-        common_limit.max_velocity = std::min(common_limit.max_velocity, it->second.max_velocity);
-      }
-      else
-      {
-        common_limit.has_velocity_limits = true;
-        common_limit.max_velocity = it->second.max_velocity;
-      }
-    }
-
-    // If this limit has acceleration limits
-    if(it->second.has_acceleration_limits)
-    {
-      // Merge if common_limit has allready limit
-      if(common_limit.has_acceleration_limits)
-      {
-        common_limit.max_acceleration = std::min(common_limit.max_acceleration, it->second.max_acceleration);
-      }
-      else
-      {
-        common_limit.has_acceleration_limits = true;
-        common_limit.max_acceleration = it->second.max_acceleration;
-      }
-    }
-
-    // If this limit has deceleration limits
-    if(it->second.has_deceleration_limits)
-    {
-      // Merge if common_limit has allready limit
-      if(common_limit.has_deceleration_limits)
-      {
-        common_limit.max_deceleration = std::max(common_limit.max_deceleration, it->second.max_deceleration);
-      }
-      else
-      {
-        common_limit.has_deceleration_limits = true;
-        common_limit.max_deceleration = it->second.max_deceleration;
-      }
-    }
-  }
-
   return common_limit;
 }
 
-pilz_extensions::JointLimit pilz::JointLimitsContainer::getLimit(const std::string &joint_name) const
+pilz_extensions::JointLimit JointLimitsContainer::getCommonLimit(const std::vector<std::string> &joint_names) const
+{
+  pilz_extensions::JointLimit common_limit;
+  for(auto joint_name : joint_names)
+  {
+    updateCommonLimit(container_.at(joint_name), common_limit);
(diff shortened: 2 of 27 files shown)


pilz_trajectory_generation/include/pilz_trajectory_generation/joint_limits_container.h
```diff
@@ -115,6 +127,15 @@ class JointLimitsContainer
   bool verifyPositionLimits(const std::vector<std::string> &joint_names,
                             const std::vector<double> &joint_positions) const;
 
+private:
+  /**
+   * @brief update the most strict limit with given joint limit
+   * @param joint_limit
+   * @param common_limit the current most strict limit
+   */
+  void updateCommonLimit(const pilz_extensions::JointLimit& joint_limit,
```
hslusarek: Function does not use any class members => make function ```static```.

pilz_trajectory_generation/src/trajectory_generator_ptp.cpp
```diff
@@ -35,37 +36,32 @@ TrajectoryGeneratorPTP::TrajectoryGeneratorPTP(const robot_model::RobotModelCons
   }
 
   joint_limits_ = planner_limits_.getJointLimitContainer();
-  mostStrictLimit_ = joint_limits_.getCommonLimit();
 
-  if(mostStrictLimit_.has_velocity_limits &&
-     mostStrictLimit_.has_acceleration_limits &&
-     mostStrictLimit_.has_deceleration_limits)
+  // collect most strict joint limits for each group in robot model
+  for(auto jmg : robot_model->getJointModelGroups())
```
hslusarek: ```const auto & jmg``` instead of copy.

pilz_trajectory_generation/test/test_utils.h
```diff
@@ -60,16 +60,15 @@ inline std::string getJointName(size_t joint_number, std::string joint_prefix)
...
    * @return
    */
-inline pilz::JointLimitsContainer createFakeLimits(size_t joint_number,
-                                                   std::string joint_prefix = testutils::JOINT_NAME_PREFIX)
+inline pilz::JointLimitsContainer createFakeLimits(robot_model::RobotModelConstPtr &robot_model)
 {
   pilz::JointLimitsContainer container;
 
-  for(size_t i = 0; i < joint_number; ++i)
+  for(std::string name : robot_model->getVariableNames())
```
hslusarek: ```const std:string& name``` better then copy of string.

pilz_trajectory_generation/src/joint_limits_container.cpp
```diff
@@ -20,141 +20,75 @@
...
+  return common_limit;
+}
 
-  common_limit = container_.begin()->second;
-
-  for(auto it = std::next(container_.begin()); it != container_.end(); ++it)
+pilz_extensions::JointLimit JointLimitsContainer::getCommonLimit(const std::vector<std::string> &joint_names) const
+{
+  pilz_extensions::JointLimit common_limit;
+  for(auto joint_name : joint_names)
```
hslusarek: ```const auto& joint_name```, to avoid copies of strings.

pilz_trajectory_generation/src/trajectory_generator_ptp.cpp
```diff
@@ -200,8 +188,18 @@ void TrajectoryGeneratorPTP::planPTP(const std::map<std::string, double>& start_
...
+      // causes the program to terminate if acc_time<=0 or dec_time<=0 (should be prevented by goal_reached block above)
+      // by using the most strict limit, the following should always return true
+      if (!velocity_profile.at(joint_name).SetProfileAllDurations(start_pos.at(joint_name), goal_pos.at(joint_name),
+                                                                  acc_time,const_time,dec_time))
+      // LCOV_EXCL_START
+      {
+        std::stringstream error_str;
+        error_str << "TrajectoryGeneratorPTP::planPTP(): Can not synchronize velocity profile of axis " << joint_name
+               << " with leading axis " << leading_axis;
+        std::runtime_error(error_str.str());
```
hslusarek: I also prefer to throw exceptions. However, I do not like the fact that this class is a mix of throwing exceptions and returning ```bool``` values to indicate success or failure. Therefore, I suggest to write an Issue stating that trajectory_generator_ptp should be refactored to use exceptions instead of ```bool``` values. Furthermore, consider using  the already existing ```TrajectoryGeneratorException``` instead of a very generic ```std::runtime_error``` exception.

pilz_trajectory_generation/test/test_utils.h
```diff
@@ -60,16 +60,15 @@ inline std::string getJointName(size_t joint_number, std::string joint_prefix)
 }
 
 /**
-   * @brief Create limits for tests to avoid the need to get the from the parameter server
-   * @param joint_number
+   * @brief Create limits for tests to avoid the need to get the limits from the parameter server
+   * @param robot_model
```
hslusarek: Please just drop the ```@param``` and ```@return``` if no documentation is given.

pilz_trajectory_generation/test/test_utils.h
```diff
@@ -60,16 +60,15 @@ inline std::string getJointName(size_t joint_number, std::string joint_prefix)
...
    * @return
    */
-inline pilz::JointLimitsContainer createFakeLimits(size_t joint_number,
-                                                   std::string joint_prefix = testutils::JOINT_NAME_PREFIX)
+inline pilz::JointLimitsContainer createFakeLimits(robot_model::RobotModelConstPtr &robot_model)
```
hslusarek: Why pass a complete ```robot_model```, if only the joint names are needed? I suggest passing: ```const std::vector<std::string>& joint_names```. Furthermore, in my opinion this function is to big to be inlined.

pilz_trajectory_generation/test/unittest_trajectory_generator_ptp.cpp
```diff
@@ -90,25 +90,24 @@ void TrajectoryGeneratorPTPTest::SetUp()
...
-  joint_limit.max_position = 3.124;
-  joint_limit.min_position = -3.124;
-  joint_limit.has_velocity_limits = true;
-  joint_limit.max_velocity = 1;
-  joint_limit.has_acceleration_limits = true;
-  joint_limit.max_acceleration = 0.5;
-  joint_limit.has_deceleration_limits = true;
-  joint_limit.max_deceleration = -1;
-  for(const auto& joint_name : joint_names)
+  for(auto jmg : robot_model_->getJointModelGroups())
```
hslusarek: Avoid copies by using ```const auto & jmg```.

pilz_trajectory_generation/test/unittest_trajectory_generator_ptp.cpp
```diff
@@ -90,25 +90,24 @@ void TrajectoryGeneratorPTPTest::SetUp()
...
   {
-    joint_limits.addLimit(joint_name, joint_limit);
+    std::vector<std::string> joint_names = jmg->getActiveJointModelNames();
```
hslusarek, line 96: Avoid copies by using ```const std::vector<std::string>& joint_names```.

pilz_trajectory_generation/test/unittest_trajectory_generator_ptp.cpp
```diff
@@ -262,7 +266,18 @@ TEST_P(TrajectoryGeneratorPTPTest, testInsufficientLimit)
   sufficient_limit.has_deceleration_limits = true;
   sufficient_limit.max_deceleration = -1;
   JointLimitsContainer sufficient_joint_limits;
-  sufficient_joint_limits.addLimit(joint_name, sufficient_limit);
+  for(auto jmg : robot_model_->getJointModelGroups())
```
hslusarek: This unit-test is very complicated to read although it tests "easy" aspects. Please consider splitting this test. For example: One test to check successful limits setting and one for insufficient limit setting.

pilz_trajectory_generation/test/unittest_trajectory_generator_ptp.cpp
```diff
@@ -222,19 +224,17 @@ TEST_P(TrajectoryGeneratorPTPTest, missingDecelerationimits)
...
  *    1. the constructor throws an exception of type TrajectoryGeneratorInvalidLimitsException
  *    2. the constructor throws no exception
  */
 TEST_P(TrajectoryGeneratorPTPTest, testInsufficientLimit)
 {
-  // joint name
-  ASSERT_TRUE(robot_model_->getJointModelGroup(planning_group_)->getJointModelNames().size())
-      << "no joint exists in the planning group.";
-  std::string joint_name = robot_model_->getJointModelGroup(planning_group_)->getJointModelNames().front();
+  auto joint_models = robot_model_->getActiveJointModels();
```
hslusarek: Avoid unnecessary copy by using ```const std::vector<JointModel*>&``` or ```const auto& joint_models```

pilz_trajectory_generation/test/unittest_trajectory_generator_ptp.cpp
```diff
@@ -262,7 +266,18 @@ TEST_P(TrajectoryGeneratorPTPTest, testInsufficientLimit)
   sufficient_limit.has_deceleration_limits = true;
   sufficient_limit.max_deceleration = -1;
   JointLimitsContainer sufficient_joint_limits;
-  sufficient_joint_limits.addLimit(joint_name, sufficient_limit);
+  for(auto jmg : robot_model_->getJointModelGroups())
+  {
+    auto joint_names {jmg->getActiveJointModelNames()};
+    ASSERT_TRUE(joint_names.size());
```
hslusarek: If you want to check if a vector is not empty, use ```joint_names.empty()```.

pilz_trajectory_generation/test/unittest_trajectory_generator_ptp.cpp
```diff
@@ -262,7 +266,18 @@ TEST_P(TrajectoryGeneratorPTPTest, testInsufficientLimit)
...
+    ASSERT_TRUE(sufficient_joint_limits.addLimit(joint_names.front(), sufficient_limit))
+      << "Failed to add the limits for joint " << joint_names.front();
+    for(auto it = std::next(joint_names.begin()); it != joint_names.end(); ++it)
```
hslusarek, line 289: Use ```std::for_each``` instead?

pilz_trajectory_generation/test/unittest_trajectory_generator_ptp.cpp
```diff
@@ -262,7 +266,18 @@ TEST_P(TrajectoryGeneratorPTPTest, testInsufficientLimit)
   sufficient_limit.has_deceleration_limits = true;
   sufficient_limit.max_deceleration = -1;
   JointLimitsContainer sufficient_joint_limits;
-  sufficient_joint_limits.addLimit(joint_name, sufficient_limit);
+  for(auto jmg : robot_model_->getJointModelGroups())
+  {
```
hslusarek, line 283: Please state in a comment what you doing within the loop because it is rather difficult to understand. If you separate the tests it might be become easier to read and understand, and then an additional comment might not be necessary.

pilz_trajectory_generation/test/unittest_trajectory_generator_ptp.cpp
```diff
@@ -205,14 +206,15 @@ TEST_P(TrajectoryGeneratorPTPTest, missingDecelerationimits)
   LimitsContainer planner_limits;
 
   pilz::JointLimitsContainer joint_limits;
-  std::vector<std::string> joint_names = robot_model_->getJointModelGroup(planning_group_)->getActiveJointModelNames();
+  auto joint_models = robot_model_->getActiveJointModels();
```
hslusarek: Avoid copy

pilz_trajectory_generation/src/trajectory_generator_ptp.cpp
```diff
@@ -163,14 +160,13 @@ void TrajectoryGeneratorPTP::planPTP(const std::map<std::string, double>& start_
...
   {
-
     // create vecocity profile if necessary
     velocity_profile.insert(std::make_pair(
                               joint_name,
                               VelocityProfile_ATrap(
-                                velocity_scaling_factor*joint_limits_.getLimit(joint_name).max_velocity,
-                                acceleration_scaling_factor*joint_limits_.getLimit(joint_name).max_acceleration,
-                                acceleration_scaling_factor*joint_limits_.getLimit(joint_name).max_deceleration)));
+                                velocity_scaling_factor * most_strict_limits_.at(group_name).max_velocity,
```
hslusarek, line 167: I don't know. This looks funny to me. If I'm the user, I expect the scaling factor to refer to the joint limit of the corresponding joint and not the group. This way I don't even know which joint will be chosen as reference for the velocity profile creation, at least not until I look specifically at all limits. We should discuss this more closely. Maybe there is only a misunderstanding on my end...

pilz_trajectory_generation/src/joint_limits_container.cpp
```diff
@@ -20,141 +20,75 @@
...
+
+bool JointLimitsContainer::addLimit(const std::string &joint_name, pilz_extensions::JointLimit joint_limit)
 {
   if(joint_limit.has_deceleration_limits && joint_limit.max_deceleration >= 0)
   {
     ROS_ERROR_STREAM("joint_limit.max_deceleration MUST be negative!");
     return false;
   }
-  else
+  if (!container_.insert(std::pair<std::string, pilz_extensions::JointLimit>(joint_name, joint_limit)).second)
```
hslusarek: It's not a good style (with regard to readability) to perform more than one operation in a one line.

hslusarek (changes requested):

pilz_trajectory_generation/src/trajectory_generator_ptp.cpp
martiniil: In my opinion we should aim for removing this block (and the throwing), since the profiles can always be computed successfully by construction.

pilz_trajectory_generation/test/test_utils.h
martiniil: Done

pilz_trajectory_generation/src/joint_limits_container.cpp
martiniil: Done

pilz_trajectory_generation/test/test_utils.h
martiniil: Done

pilz_trajectory_generation/src/trajectory_generator_ptp.cpp
martiniil: Done

pilz_trajectory_generation/include/pilz_trajectory_generation/joint_limits_container.h
martiniil: Done

pilz_trajectory_generation/test/test_utils.h
martiniil: Done

pilz_trajectory_generation/test/unittest_trajectory_generator_ptp.cpp
martiniil: Done

pilz_trajectory_generation/test/unittest_trajectory_generator_ptp.cpp
martiniil: Done

pilz_trajectory_generation/test/unittest_trajectory_generator_ptp.cpp
martiniil: Done

pilz_trajectory_generation/test/unittest_trajectory_generator_ptp.cpp
martiniil: Done

pilz_trajectory_generation/test/unittest_trajectory_generator_ptp.cpp
martiniil: This doesn't work, since  sufficient_joint_limits is modified in the loop.

pilz_trajectory_generation/test/unittest_trajectory_generator_ptp.cpp
martiniil: Done

pilz_trajectory_generation/test/unittest_trajectory_generator_ptp.cpp
martiniil: I would suggest not to split it in order to avoid code duplication. I will restructure it and add comments in order to make the steps more clear. Then you can decide if you like it.

pilz_trajectory_generation/test/unittest_trajectory_generator_ptp.cpp
martiniil: Done

pilz_trajectory_generation/src/joint_limits_container.cpp
martiniil: Done. Please review changes

pilz_trajectory_generation/src/trajectory_generator_ptp.cpp
martiniil: As discussed we keep it according to our ReadMe

pilz_trajectory_generation/src/trajectory_generator_ptp.cpp
martiniil: see #134

hslusarek (approved):

PR #133, opened 2019-04-03
Add missing pilz_msgs to catkin_package()

hslusarek:


jschleicher (approved):
yep, thanks. Please merge after Travis goes green.

gavanderhoorn:
Just curious: you have `catkin_lint` enabled as part of the CI ([here](https://github.com/PilzDE/pilz_industrial_motion/blob/5dc82a3129b557991ff4ef45a2bbbce5324cf43f/.travis.yml#L18)). Did it not catch this?

hslusarek:
I was also wondering about that. Actually the problem was found, see [here](https://travis-ci.org/PilzDE/pilz_industrial_motion/jobs/515167945) (line 605).  I think the CI did not fail because it is only stated as `warning`.

gavanderhoorn:
Could consider adding `--strict` (treat warnings as errors).

hslusarek:
Thanks. Let's give this a try (#135).

Issue #102, opened 2019-02-22
prbt_base is not default reference_frame


martiniil:
related: #98

### Steps to reproduce
1. Modify urdf such that `prbt_base` does not match root-link `world`
2. Execute a ptp pose command without reference frame

### Expected behavior
Robot should move relative to `prbt_base` frame

### Observed behavior
Robot moves relative to `world` frame

martiniil:
Is it intended that we require `prbt_base` and `world` to be identical? I am not sure if I fully understand the meaning of these frames. At least it should be possible to position the robot by simply altering the transform from `world` to `prbt_base_link`?
![screenshot from 2019-02-22 14-47-53](https://user-images.githubusercontent.com/35950815/53246426-f25f8100-36b0-11e9-8e08-97bc6b49706f.png)

jschleicher:
Please see [ROS-I wiki: Create URDF for an Industrial Robot](http://wiki.ros.org/Industrial/Tutorials/Create%20a%20URDF%20for%20an%20Industrial%20Robot#Additional.2BAC8-Standard_Frames) for a description of standard frames.

Yes, it is possible to move the robot around / move prbt_base to not match the world coordinate system. If you want to command positions in "world" or "base" coordinates in the API is a question of conventions. I'm not aware of a standard there.

[MoveIt commander](http://docs.ros.org/kinetic/api/moveit_tutorials/html/doc/move_group_python_interface/move_group_python_interface_tutorial.html#planning-to-a-pose-goal) refers to a group for every motion plan request; I think that sounds like a good default. Maybe you should choose the default reference frame based on the default planning group as well?

gavanderhoorn:
> Please see [ROS-I wiki: Create URDF for an Industrial Robot](http://wiki.ros.org/Industrial/Tutorials/Create%20a%20URDF%20for%20an%20Industrial%20Robot#Additional.2BAC8-Standard_Frames) for a description of standard frames

Slightly off-topic: a more up-to-date (but still WIP) reference: [Coordinate Frames for Serial Industrial Manipulators](https://github.com/gavanderhoorn/rep/blob/coord_frames_ind_robots/rep-xxxx.rst) ([rendered](http://gavanderhoorn.github.io/rep/rep-xxxx.html)).

> Yes, it is possible to move the robot around / move prbt_base to not match the world coordinate system. If you want to command positions in "world" or "base" coordinates in the API is a question of conventions. I'm not aware of a standard there.

[...]

PR #141, opened 2019-04-08
Fix/default ref frame

martiniil:
Targets #102
- [x] update changelog
- [x] increase API version

### DIFF
--- a/pilz_robot_programming/src/pilz_robot_programming/commands.py
+++ b/pilz_robot_programming/src/pilz_robot_programming/commands.py
@@ -273,17 +273,17 @@ class _BaseCmd(_AbstractCmd):
         """Determines the goal pose for the given command."""
         current_pose = robot.get_current_pose(target_link=self._target_link, base=self._reference_frame)
 
-        if self._relative:
-            self._goal = _pose_relative_to_absolute(current_pose, self._goal)
-
-        if not self._reference_frame == _DEFAULT_BASE_LINK:
-            return _to_robot_reference(robot, self._reference_frame, self._goal)
+        goal_pose = self._goal
 
         # in case of uninitialized orientation, set the goal orientation as current
-        if _is_quaternion_initialized(self._goal.orientation):
-            return self._goal
-        else:
-            return Pose(position=self._goal.position, orientation=current_pose.orientation)
+        identity_quat = Quaternion(w=1.0)
+        if not _is_quaternion_initialized(goal_pose.orientation):
+            goal_pose.orientation = identity_quat if self._relative else current_pose.orientation
+
+        if self._relative:
+            goal_pose = _pose_relative_to_absolute(current_pose, goal_pose)
+
+        return _to_robot_reference(robot, self._reference_frame, goal_pose)
 
     def _get_joint_pose(self, robot):
         """Determines the joint goal for the given command."""
@@ -649,9 +649,6 @@ def _to_robot_reference(robot, pose_frame, goal_pose_custom_ref):
     assert isinstance(goal_pose_custom_ref, Pose)
 
     robot_ref = robot._robot_commander.get_planning_frame()
-
-    if not _is_quaternion_initialized(goal_pose_custom_ref.orientation):
-        goal_pose_custom_ref.orientation.w = 1
 
     if pose_frame == robot_ref:
         return goal_pose_custom_ref
--- a/pilz_robot_programming/test/integrationtests/tst_api_cmd_conversion.py
+++ b/pilz_robot_programming/test/integrationtests/tst_api_cmd_conversion.py
@@ -29,6 +29,7 @@ from pilz_robot_programming.commands import *
 _TEST_DATA_FILE_NAME = RosPack().get_path("pilz_industrial_motion_testutils") + "/test_data/testdata_deprecated.xml"
 PLANNING_GROUP_NAME = "manipulator"
 TARGET_LINK_NAME = "prbt_tcp"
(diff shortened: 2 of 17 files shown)


pilz_robot_programming/src/pilz_robot_programming/commands.py
```diff
@@ -273,17 +273,18 @@ def _get_goal_pose(self, robot):
...
-        if not self._reference_frame == _DEFAULT_BASE_LINK:
-            return _to_robot_reference(robot, self._reference_frame, self._goal)
+        goal_pose = self._goal
 
         # in case of uninitialized orientation, set the goal orientation as current
-        if _is_quaternion_initialized(self._goal.orientation):
-            return self._goal
-        else:
-            return Pose(position=self._goal.position, orientation=current_pose.orientation)
+        identity_quat = Quaternion()
```
gavanderhoorn:
```suggestion
        identity_quat = Quaternion(w=1.0)
```

martiniil: Thanks! @gavanderhoorn

jschleicher (commented):
Where is the actual change? I cannot see from the diff, where this fixes #102. Could you explain, please?

What about documentation and changelog? Does this change the behaviour and make a `ROBOT_API ` change necessary?

martiniil:
These lines introduced the bug:

        if not self._reference_frame == _DEFAULT_BASE_LINK:
            return _to_robot_reference(robot, self._reference_frame, self._goal)

The actual change is that now the goal is always transformed to the root frame of the robot.

martiniil:
It does change the behaviour of the API. Now it behaves as promised by the API docu.

pilz_robot_programming/src/pilz_robot_programming/commands.py
```diff
@@ -640,9 +640,6 @@ def _to_robot_reference(robot, pose_frame, goal_pose_custom_ref):
 
     robot_ref = robot._robot_commander.get_planning_frame()
 
-    if not _is_quaternion_initialized(goal_pose_custom_ref.orientation):
-        goal_pose_custom_ref.orientation.w = 1
-
```
jschleicher, line 655: Would it be feasible to delegate this task to moveit? Simply set the `header.frame_id` in your request and leave the transformation to the motion planner. As far as I understand, it should be implemented there (as well) and thus is some code duplication...

jschleicher:
Okay, citing @agutenkunst (see [#102-issuecomment-467816895](https://github.com/PilzDE/pilz_industrial_motion/issues/102#issuecomment-467816895)) this requires changing the `ROBOT_API`, since the custom python code could move the robot differently after this change.

pilz_robot_programming/src/pilz_robot_programming/commands.py
martiniil: Sounds good. I will give it a try.

gavanderhoorn: > Would it be feasible to delegate this task to moveit?

While in general I would agree with reusing as much as possible, in this case it might not be such a bad thing to try and make sure that things under control of the planner / robot programming framework are consistent.

It's a small thing to make sure that all outgoing planning requests / trajectories / poses are in robot base frame. It also makes the software in this repository more robust against changes in MoveIt (it might stop doing what @jschleicher writes in his comment) and if there is a planning framework that is not MoveIt, but provides the same/a similar interface, but does not transform poses into robot base frame, bad things can happen.

I'm a rather defensive programmer, so this may not be compatible with the style of the rest of `pilz_robot_programming`.

martiniil: @jschleicher @gavanderhoorn Thanks for your inputs. I don't have a strong opinion about it. I would keep it as it is, unless more reviewers request a change.

agutenkunst: @gavanderhoorn thank you for your input!
I also agree with @jschleicher that the motion planner should be able to handle other frame_ids but maybe we should implement this independent from the need of the underlying PythonAPI.

pilz_robot_programming/test/integrationtests/tst_api_cmd_conversion.py
```diff
@@ -29,7 +29,8 @@
 _TEST_DATA_FILE_NAME = RosPack().get_path("pilz_industrial_motion_testutils") + "/test_data/testdata_deprecated.xml"
 PLANNING_GROUP_NAME = "manipulator"
 TARGET_LINK_NAME = "prbt_tcp"
-API_VERSION = "1"
+BASE_LINK_NAME = "prbt_base"
+API_VERSION = "2"
```
agutenkunst, line 33: I would like to open the discussion if we support backwards compatibility to some degree...
Any opinion? To much effort?

jschleicher: +1 for allowing execution of API=1 programs (with former behaviour) and a deprecation warning

pilz_robot_programming/src/pilz_robot_programming/commands.py
jschleicher: Ok, fine. Keep this for now.

jschleicher:
@agutenkunst  @martiniil any progress?

martiniil:
@jschleicher We discussed several concepts for a backwards compability and scheduled its implementation. Until then, merging this PR is postponed.

PR #145, opened 2019-04-12
reduce prbt_support to `<test_depend>`

jschleicher:
The debian package pilz-trajectory-generation currently depends on prbt_support; that should not be necessary.
A test depend should suffice?!


agutenkunst:
Sounds reasonable. Thank you!

agutenkunst (approved):

agutenkunst:
...

PR #150, opened 2019-04-17
Switch active devel branch: kinetic->melodic

martiniil:
Updates ReadMe files

After this change I suppose that we simply open any new PR for melodic-devel. We can keep the old ones and merge one last time from kinetic->melodic after the next release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,8 +13,8 @@ The FTP is implemented by Pilz GmbH & Co. KG.
 
 
 ### Note to Developers:
-`kinetic-devel` is considered to be the active development branch.
-Changes are merged into `melodic-devel` with every release or on demand.
+`melodic-devel` is considered to be the active development branch.
+Relevant changes are cherry-picked into `kinetic-devel` on a case-by-case basis.
 
 ## Package: pilz_trajectory_generation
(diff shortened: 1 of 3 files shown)


ct2034 (commented):
I think we also need to change it in the repo settings. But probably only @dbakovic can do it.

jschleicher:
And you probably want to have a branching model explained: E.g. do a final merge of kinetic into melodic *before merging this PR*. From then on cherry-pick relevant PRs into kinetic on a case-by-case basis ("backport")? Backwards-merge (as stated in the Readme) is not a use-case, I've seen anywhere else...

martiniil:
@jschleicher I would rather do a final kinetic->melodic merge _after_ merging this PR.

@ct2034 Which settings would you change?

ct2034:
You can change it like this (if you have the correct permissions): https://help.github.com/en/articles/setting-the-default-branch

martiniil:
@ct2034 Can we separate the change in the github settings from this PR?

martiniil:
@PilzDE/robotics  Can we merge this and release afterwards?

hslusarek (approved):

PR #153, opened 2019-04-29
WIP: Fix end-effector assumption in getTipFrame function

hslusarek:
The name of the function and the actual functionality of the getTipFrame
function was confusing. Therefore, the function name and the functionality
was adapted. Depending code was changed accordingly.

### DIFF
--- a/pilz_trajectory_generation/include/pilz_trajectory_generation/tip_frame_getter.h
+++ b/pilz_trajectory_generation/include/pilz_trajectory_generation/tip_frame_getter.h
@@ -1,8 +1,27 @@
+/*
+ * Copyright (c) 2019 Pilz GmbH & Co. KG
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
 #ifndef TIP_FRAME_GETTER_H
 #define TIP_FRAME_GETTER_H
 
 #include <string>
 #include <cassert>
+#include <stdexcept>
+#include <vector>
 
 #include <moveit/robot_model/joint_model_group.h>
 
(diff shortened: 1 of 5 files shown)


pilz_trajectory_generation/test/integrationtest_command_list_manager.cpp
```diff
@@ -629,7 +629,7 @@ TEST_F(IntegrationTestCommandListManager, TestGroupSpecificStartState)
 TEST_F(IntegrationTestCommandListManager, TestTipFrameNoEndEffector)
 {
   Gripper gripper_cmd {data_loader_->getGripper("open_gripper")};
-  EXPECT_THROW(getTipFrame(robot_model_->getJointModelGroup(gripper_cmd.getPlanningGroup())), EndEffectorException);
```
martiniil, line 632: I think `EndEffectorException` is not needed any more, so we can remove it completely.

pilz_trajectory_generation/src/command_list_manager.cpp
```diff
@@ -108,7 +108,8 @@ bool CommandListManager::checkRadiiForOverlap(const robot_trajectory::RobotTraje
     return false;
   }
 
-  const std::string& blend_frame {getTipFrame(model_->getJointModelGroup(traj_A.getGroupName()))};
+  assert(!model_->getJointModelGroup(traj_A.getGroupName())->isEndEffector());
```
jschleicher: Thanks for your follow-up.

Is this assertion necessary? You already check for the existence of a solver in https://github.com/PilzDE/pilz_industrial_motion/blob/3f6150551143d83f9ea30fe883f99ce590375c4e/pilz_trajectory_generation/include/pilz_trajectory_generation/tip_frame_getter.h#L25. Why is an end-effector not allowed here? The PG+70 does not have a solver defined an is thus excluded.

martiniil (changes requested):

pilz_trajectory_generation/src/command_list_manager.cpp
martiniil: You are right, It is not necessary.
Imho it would be nice, if the user obtains an error message like e.g. "blending of an end-effector movement is not supported."

jschleicher: Indeed a Warning / Error message is more user-friendly than an assertion.

But on the other hand: What would be necessary, to allow end-effector-blending for groups that have a solver? Is there anything missing or it this just a defensive "we've not tested it"?

hslusarek: I think we have a huge misunderstanding when it comes to "assertions". So let's just talk about it for a second:
Assertions are things that should never ever happen. However, I might be in a function where I assume that a certain assumptions holds, which, for example, is checked beforehand. To ensure that the programmer reading the functions knows about this implicit assumption I add an assertion. Also, I want to ensure that the conditions always holds, even if somebody accidentally deletes the check which should be performed beforehand.

Now, back to the concrete problem:
Yes, I actually agree with both of you and more importantly the check and the warning already exist [here](https://github.com/PilzDE/pilz_industrial_motion/blob/bd5a71ccb64559fbd050525b761af6b7afa70fb0/pilz_trajectory_generation/src/command_list_manager.cpp#L181) and [here](https://github.com/PilzDE/pilz_industrial_motion/blob/bd5a71ccb64559fbd050525b761af6b7afa70fb0/pilz_trajectory_generation/src/command_list_manager.cpp#L196). However, like stated before, this is not the reason for the assertion. The assertion exists for the next programmer reading this lines and for the absolute worst case scenario that somebody accidentally deletes the end-effector check beforehand.

@jschleicher 
A pure end-effector blending does not make much sense for me. Why should I want to blend a gripper open or close movement?
The only exception I can think of is the case where you have a group consisting of a manipulator + end-effector. But this case should work because, in that case, the group is not a pure end-effector.

martiniil: Thanks @hslusarek. I missed the check you mention.

martiniil: It seems one can have different opinion whether a blending of pure end-effector groups makes sense or not. Regardless, why should we categorically exclude it, despite our code is capable of it?

I suggest that we do not care, if a group is an end-effector [here](https://github.com/PilzDE/pilz_industrial_motion/blob/bd5a71ccb64559fbd050525b761af6b7afa70fb0/pilz_trajectory_generation/src/command_list_manager.cpp#L181), but only require that a solver exists. Why not? Then the asserts can be omitted definitely.

jschleicher: End-effectors can be multi-joint groups (like articulated hands) and there blending is as useful as it is for the serial manipulator.

pilz_trajectory_generation/test/integrationtest_command_list_manager.cpp
hslusarek: Done.

pilz_trajectory_generation/src/command_list_manager.cpp
hslusarek: Ok. I replaced the assumption that end-effector cannot be blended with the more general assumptions that only groups without solver cannot be blended (see [here](https://github.com/PilzDE/pilz_industrial_motion/blob/cfcee3bfb85ae29a09e690960aba19453e71c6a3/pilz_trajectory_generation/src/command_list_manager.cpp#L179)). However, solver allow more than one tip frame. What should we do in this case? Simply take the first one? Return with error?

pilz_trajectory_generation/include/pilz_trajectory_generation/tip_frame_getter.h
```diff
@@ -11,27 +12,36 @@
...
+}
 
-  auto solver {group->getSolverInstance()};
-  if(solver == nullptr)
+/**
+ * @return The name of the tip frame (link) of the specified group
+ * returned by the solver. An exception is thrown if no solver for
+ * the group could be found.
+ *
+ * @throws exception in case the solver has no solver.
```
jschleicher:
```suggestion
 * @throws exception in case the group has no solver.
```

pilz_trajectory_generation/include/pilz_trajectory_generation/tip_frame_getter.h
```diff
@@ -11,27 +12,36 @@
...
+ */
+static const std::string& getSolverTipFrame(const moveit::core::JointModelGroup* group)
+{
+  if( !hasSolver(group) )
   {
     throw NoSolverException("No solver for group " + group->getName());
   }
-  return solver->getTipFrame();
+  return group->getSolverInstance()->getTipFrame();
```
jschleicher: Yes, I'd suggest to allow only planning groups with a single tipFrame.
You could use the convenience function `getOnlyOneEndEffectorTip()` (which does not throw, but returns a nullptr instead):
```suggestion
  auto eegroup = group->getOnlyOneEndEffectorTip();
  return eegroup->getName();
```

jschleicher (commented):
Yes, I'd agree to allow only planning groups with a single tipFrame and output an error message / throw an exception otherwise.

jschleicher:
@hslusarek ping

martiniil (approved):

PR #157, opened 2019-05-09
added methods for 'with' usage to Robot

rfeistenauer:
Since the only save way, other than deleting the node, to free the Robot instance is to call its deconstructor "__del__()" id suggest to add the context_guard.

With it the Robot can savely be used for some motion and freed again within a node like:
```
with Robot('1') as rob:
   rob.move(...)

r = Robot('1') # would work here
```

This is relevant to switch robot control between several nodes.


hslusarek:
Please continue on PR #159

PR #158, opened 2019-05-10
Python api methods for brake test support

ct2034:
Related: https://github.com/PilzDE/pilz_robots/pull/120

This  is wip. A couple of open points:

- [x] ~~US-20251 says "Prüfe, ob ein Bremsentest durchgeführt wird" - How should we do this? How is the BrakeTest srv intended? I thought it will execute the test and return the result. But the result values look like actionlib could work better, here. -> TK-75819~~
- [x] There are TODOs in the files (in `robot.py` and `tst_api_brake_test.py`)
- [x] Probably more integration tests are needed
- [x] Check for same service names / make them configurable
- [x] Should we also have asynchronous versions of the api calls? Now the block ...


martiniil:
@ct2034 wip still?

pilz_robot_programming/doc/README.md
```diff
@@ -18,3 +18,14 @@ The move control request "tells" the thread executing the move function:
 * to perform normally.  
 
 ![Move orders state diag](diag_state_move_orders.png)
+
+# Brake test
```
hslusarek, line 22: I think we should consider writing down more details why a brake test is required and what it is. I'm not sure if we want to do it in this story or another one?

pilz_robot_programming/src/pilz_robot_programming/robot.py
```diff
@@ -16,23 +16,24 @@
...
 
+from pilz_msgs.msg import MoveGroupSequenceAction
+from prbt_hardware_support.srv import IsBrakeTestRequired, BrakeTest, BrakeTestResponse
 from .move_control_request import _MoveControlState, MoveControlAction,_MoveControlStateMachine
 from .commands import _AbstractCmd, _DEFAULT_PLANNING_GROUP, _DEFAULT_TARGET_LINK, _DEFAULT_BASE_LINK
 from .exceptions import *
-from geometry_msgs.msg import Quaternion, PoseStamped, Pose
-from std_msgs.msg import Header
 
 __version__ = '1.0.0'
```
hslusarek: Increase minor number to account for new functionality?

pilz_robot_programming/src/pilz_robot_programming/robot.py
```diff
@@ -274,6 +280,63 @@ def resume(self):
...
+                IsBrakeTestRequired)
+            resp = is_brake_test_required_client()
+            if resp.result:
+                rospy.loginfo("Brake Test REQUIRED")
+            else:
+                rospy.loginfo("Brake Test NOT REQUIRED")
+            return resp.result
+        except rospy.ROSException, e:
+            rospy.logerr("Service call failed: {0}".format(e))
+            raise e
```
hslusarek, line 308: In other cases, we caught the exception like you do here but instead of throwing the same exception, we introduced our own exception, in your case `RobotBrakeTestException`. This leads to more understandable error messages on the console (at least that's the idea :) ) because you can see that the current exception has something to do with the brake test.

pilz_robot_programming/src/pilz_robot_programming/robot.py
```diff
@@ -274,6 +280,63 @@ def resume(self):
...
+                resp.success,
+                resp.error_msg
+            ))
+            if not resp.success:
+                e = RobotBrakeTestException(resp.error_code, resp.error_msg)
+                rospy.logerr("Brake Test returned: " + str(e))
+                raise e
+        except rospy.ROSException, e:
+            rospy.logerr("Service call failed: {0}".format(e))
+            raise e
```
hslusarek, line 345: See my comment from the `is_brake_test_required()` function.

pilz_robot_programming/src/pilz_robot_programming/robot.py
```diff
@@ -274,6 +280,63 @@ def resume(self):
...
+            Function blocks until brake test is finished.
+        """
+        rospy.loginfo("Executing brake test")
+        try:
+            rospy.wait_for_service(self._BRAKE_TEST_EXECUTE_SRV, self._SERVICE_WAIT_TIMEOUT_S)
+            execute_brake_test_client = rospy.ServiceProxy(
+                self._BRAKE_TEST_EXECUTE_SRV,
+                BrakeTest
+            )
+            resp = execute_brake_test_client()
```
hslusarek, line 326: It is actually bad style to create try/catch blocks which contain more code than absolutely necessary (buzzword: clean code). Looking [here](http://wiki.ros.org/rospy/Overview/Services#Service_proxies) and [here](http://docs.ros.org/api/rospy/html/rospy.impl.tcpros_service.ServiceProxy-class.html#call), it seems to me that actually only the direct service calls needs a try/catch. Therefore, I suggest to down size the try/catch block.

pilz_robot_programming/src/pilz_robot_programming/robot.py
```diff
@@ -274,6 +280,63 @@ def resume(self):
...
+            rospy.loginfo("Brake Test Success: {0:b}, msg: {1}".format(
+                resp.success,
+                resp.error_msg
+            ))
+            if not resp.success:
+                e = RobotBrakeTestException(resp.error_code, resp.error_msg)
+                rospy.logerr("Brake Test returned: " + str(e))
+                raise e
```
hslusarek: I suggest throwing your own exception outside the try/catch block. It is, on the one hand, very confusing to read and, on the other hand, can easily lead to errors in the future, in case the code is extended/modified by somebody else. And yes, this includes that the variable `resp` needs to be initialized outside the try/catch block, before the `execute_brake_test_client()` function is called.

pilz_robot_programming/test/acceptance_tests/acceptance_test_brake_test.md
```diff
@@ -0,0 +1,44 @@
...
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+-->
+
+# Acceptance Test for brake test using Python-API and real robot
```
hslusarek, line 18: Looking at the `*.md` files and the corresponding `*.py` files for acceptance tests, it seems to me that both files are highly redundant. To me it seems that, in case, we have a `*.py` file for an acceptance test, we do not really need a corresponding `*.md*` file. Everything what is written in the `*.md` file could easily be stated in the `*.py` file. I also shortly talked with @martiniil and he agreed. So my suggestion: Delete the `*.md` file and copy information from `*.md` file into `*.py` file.

pilz_robot_programming/test/integrationtests/tst_api_brake_test.py
```diff
@@ -0,0 +1,134 @@
...
+        mock.start()
+        mock.advertise_brake_test_required_service()
+
+        mock.set_is_brake_test_required(False)
+        res = None
+        try:
+            res = self.robot.is_brake_test_required()
+        except Exception as e:
+            rospy.logerr(e)
+            raise e
```
hslusarek: If an exception is raised here, what happens with the thread of the `BrakeTestMock`? Might it be better to construct, start and stop the `BrakeTestMock` using the `setUp` and `tearDown` function?

pilz_robot_programming/test/integrationtests/tst_api_brake_test.py
```diff
@@ -0,0 +1,134 @@
...
+        mock.start()
+        mock.advertise_brake_test_required_service()
+
+        mock.set_is_brake_test_required(True)
+        res = None
+        try:
+            res = self.robot.is_brake_test_required()
+        except Exception as e:
+            rospy.logerr(e)
+            raise e
```
hslusarek: If everything is ok, no exception should be thrown, so why catch it here? Is it not easier and better (in this case at least) to simply ignore possible exceptions? If an exception raises, it will be automatically forwarded to the console.

pilz_robot_programming/examples/demo_brake_test_program.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python
```
hslusarek, line 1: catkin_lint reports:
```
pilz_robot_programming: warning: file 'examples/demo_brake_test_program.py' is executable but not installed
```
=> Add `demo_brake_test_program.py` to `CMakeLists.txt` [here](https://github.com/PilzDE/pilz_industrial_motion/blob/4e311ce223ce6c4a1bb3c3e2bb2299970ecfcf67/pilz_robot_programming/CMakeLists.txt#L14).

pilz_robot_programming/test/integrationtests/tst_api_brake_test.py
```diff
@@ -0,0 +1,134 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import unittest
+
+import rospy
+
+from pilz_robot_programming.exceptions import RobotBrakeTestException
+from pilz_robot_programming.robot import Robot
+from brake_test_mock import BrakeTestMock
+from prbt_hardware_support.srv import BrakeTestResponse
```
hslusarek: Import not used => Delete

hslusarek (dismissed):

pilz_robot_programming/doc/README.md
agutenkunst: I think this should belong way more high level where the whole concept is explained.
At the api level the documentation of the pure functionality should be sufficient.

pilz_robot_programming/src/pilz_robot_programming/robot.py
agutenkunst: @ct2034 @hslusarek @martiniil should we keep this __version__ in sync with the package version?

pilz_robot_programming/test/integrationtests/tst_api_brake_test.py
agutenkunst: Resolved by putting Mock shutdown into the except

pilz_robot_programming/test/integrationtests/tst_api_brake_test.py
agutenkunst: Removed except

pilz_robot_programming/src/pilz_robot_programming/robot.py
agutenkunst: Fixed.

pilz_robot_programming/test/integrationtests/tst_api_brake_test.py
agutenkunst: Thread gets joined by going out of scope...

pilz_robot_programming/test/integrationtests/tst_api_brake_test.py
agutenkunst: Removed

pilz_robot_programming/examples/demo_brake_test_program.py
agutenkunst: Done

pilz_robot_programming/src/pilz_robot_programming/robot.py
agutenkunst: Done

pilz_robot_programming/test/acceptance_tests/acceptance_test_brake_test.md
agutenkunst: I am a bit more indifferent on this topic but would suggest to discuss this for similiar tests in the future.

pilz_robot_programming/src/pilz_robot_programming/robot.py
agutenkunst: I increased the version btw. But still like to discuss the point above...

agutenkunst (approved):

agutenkunst (approved):

PR #159, opened 2019-05-13
Feature/context guard for robot

hslusarek:
Melodic version of PR #157


hslusarek:
@rfeistenauer Please have a look. I added a test with a for-loop which shows that the suggested commit does not work.

rfeistenauer:
The problem is not the with statement, but the usage of 'del rob' in previous tests.
The instance is not freed soon enough.
If you use `rob.__del__()` instead, the tests pass.

jschleicher:
@hslusarek @rfeistenauer For future PRs: You can change the target / base branch by pressing "Edit" in the top right...

hslusarek:
@jschleicher Hm... I can only change the title.

PR #162, opened 2019-05-16
Drop unused variables in Python API

### LINKED ISSUE #124: pilz_robot_programming has a lot of unused code
### Commit
commit 685c781874d1771b67058b040983813d9185a719

### Steps to reproduce
vulture .../pilz_robot_programming

### Expected behavior
Not much unused stuff

### Observed behavior
quite the opposite. And it doesn't even find all. e.g. _MAX_ACC_SCALE

jschleicher:
Fixes #124.

Maximum acceleration  is unused but probably should be re-considered when increasing the version (allowing breaking changes). Drop RobotMoveInvalidState exception and other unused imports.

### DIFF
--- a/pilz_robot_programming/src/pilz_robot_programming/commands.py
+++ b/pilz_robot_programming/src/pilz_robot_programming/commands.py
@@ -21,26 +21,24 @@ import rospy
 from tf import transformations
 from geometry_msgs.msg import Quaternion, Pose
 from geometry_msgs.msg import PoseStamped
-from pilz_msgs.msg import MoveGroupSequenceGoal, MotionSequenceRequest, MotionSequenceItem
+from pilz_msgs.msg import MoveGroupSequenceGoal, MotionSequenceItem
 from moveit_msgs.msg import (OrientationConstraint, MotionPlanRequest, JointConstraint, Constraints,
-                             MoveGroupGoal, PositionConstraint, PlanningOptions)
+                             PositionConstraint, PlanningOptions)
 import shape_msgs.msg as shape_msgs
 from operator import add
 from math import pi
 
 from .move_control_request import _MoveControlState
-from .exceptions import RobotCurrentStateError
 from copy import deepcopy
 
 
 __version__ = '0.0.dev1'
 
 # Default velocities
-_MAX_VEL_SCALE = 1
-_DEFAULT_VEL_SCALE = 0.1
+_DEFAULT_CARTESIAN_VEL_SCALE = 0.1
+_DEFAULT_JOINT_VEL_SCALE = 1.0
 
-# Default accelerations
-_MAX_ACC_SCALE = 1
+# Default acceleration
 _DEFAULT_ACC_SCALE = 0.1
 
 # Tolerance for cartesian pose
@@ -176,7 +174,7 @@ class _BaseCmd(_AbstractCmd):
     :type reference_frame: string
     """
     def __init__(self, goal=None, planning_group=_DEFAULT_PLANNING_GROUP, target_link=_DEFAULT_TARGET_LINK,
-                 vel_scale=_DEFAULT_VEL_SCALE, acc_scale=_DEFAULT_ACC_SCALE, relative=False,
+                 vel_scale=_DEFAULT_CARTESIAN_VEL_SCALE, acc_scale=_DEFAULT_ACC_SCALE, relative=False,
                  reference_frame=_DEFAULT_BASE_LINK, *args, **kwargs):
         super(_BaseCmd, self).__init__(*args, **kwargs)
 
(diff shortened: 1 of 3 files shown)


rfeistenauer:
removed the unused imports

pilz_robot_programming/src/pilz_robot_programming/robot.py
```diff
@@ -21,14 +21,14 @@
...
-from moveit_msgs.msg import MoveItErrorCodes, MoveGroupAction
+from moveit_msgs.msg import MoveItErrorCodes
 import time
 import threading
 from std_srvs.srv import Trigger
 import tf
 import psutil
 
-from .move_control_request import _MoveControlState, MoveControlAction,_MoveControlStateMachine
+from .move_control_request import _MoveControlState, MoveControlAction, MoveControlStateMachine
```
jschleicher: why did you remove the underscore?

jschleicher (commented):
@rfeistenauer thanks. I'm just fixing-up the tests to run full-speed again...

pilz_robot_programming/src/pilz_robot_programming/robot.py
rfeistenauer: If it was me then maybe on accident while deleting the unused modules?

rfeistenauer (approved):
Despite the fact, that the test changes show, that there maybe should be a general speed option, everything seems to be ok.

martiniil:
@jschleicher Changing the default acceleration changes the robots behaviour on existing programs. Why do you think it was a typo?

jschleicher:
@martiniil I think, the PTP init line should read default=default to be consistent with Lin and Circ constructors:
https://github.com/PilzDE/pilz_industrial_motion/blob/47dfa2561c14b5b187eadb207d977a93236b782e/pilz_robot_programming/src/pilz_robot_programming/commands.py#L368
https://github.com/PilzDE/pilz_industrial_motion/blob/47dfa2561c14b5b187eadb207d977a93236b782e/pilz_robot_programming/src/pilz_robot_programming/commands.py#L429

and with the default acceleration in the base class
https://github.com/PilzDE/pilz_industrial_motion/blob/47dfa2561c14b5b187eadb207d977a93236b782e/pilz_robot_programming/src/pilz_robot_programming/commands.py#L174-L176

jschleicher:
Rebased onto current melodic_devel to resolve merge-conflicts and squashed into a single commit.

martiniil:
@jschleicher Reducing the default velocity requires a version increment. Cannot merge this at the moment.

hslusarek:
> See #124.
> 
> The full velocity for PTP commands was a typo?! Maximum velocity and acceleration are then unused and can be dropped. RobotMoveInvalidState exception as well.

@jschleicher @martiniil I think a higher default velocity  for the PTP was not a typo. In my opinion we want the commands to be executed as fast as possible. However, in case of lin, circ and blending a high default velocity was not a good choice because many motions could not be performed without explicitly changing/reducing the velocity. Therefore, the default velocity was reduced. I think the general behavior should be decided by @dbakovic

pilz_robot_programming/src/pilz_robot_programming/commands.py
```diff
@@ -320,7 +317,7 @@ class Ptp(_BaseCmd):
 
             acc_scale = vel_scale * vel_scale
     """
-    def __init__(self, vel_scale=_MAX_VEL_SCALE, acc_scale=None, *args, **kwargs):
+    def __init__(self, vel_scale=_DEFAULT_VEL_SCALE, acc_scale=None, *args, **kwargs):
```
hslusarek: Please do not change to such a low default velocity. Choose a "reasonable" default velocity for PTP. This low default value was mainly intended for lin, circ and blending.

hslusarek (dismissed):

pilz_robot_programming/src/pilz_robot_programming/commands.py
ct2034: @jschleicher

jschleicher: Renamed the variable without changing its behaviour. Having different scales still seems like a typo to me; resolved into a new issue (related to operation modes?).

jschleicher: Resolved into #191.

jschleicher:
1422a95  reduces this PR to only drop unused variables. @ct2034 Please re-review.

Leaving default speed as open issue #191.

jschleicher:
@ct2034 friendly ping

ct2034 (approved):
These changes make sense to me

PR #167, opened 2019-05-22
remove misplaced changelog entry

martiniil:
Bloom-release told me that this is a mistake, therefore fixing it before release.


jschleicher (approved):
Obviously slipped in from the melodic branch.

Remove in kinetic.

PR #171, opened 2019-05-27
check for link name during request validation

jschleicher:
instead of spitting thousands of ERRORS during intersection point search

Trying to lookup the transform spits an error but does not throw, so the intersection search goes through the whole trajectory: [See Travis jobs](https://travis-ci.org/PilzDE/pilz_industrial_motion/jobs/537648379#L7483-L8134)

This PR checks for existance of the link in advance during request validation.

- [x]  Check line coverage:
  This PR does not change the coverage. Previously uncovered and still uncovered lines (copy into new issue?):
  https://github.com/PilzDE/pilz_industrial_motion/blob/5a1bd1c2d2d6d5f3a2c9aa3156bd097b49f0eadd/pilz_trajectory_generation/src/trajectory_generator_lin.cpp#L160-L162

### DIFF
--- a/pilz_trajectory_generation/src/trajectory_blender_transition_window.cpp
+++ b/pilz_trajectory_generation/src/trajectory_blender_transition_window.cpp
@@ -135,6 +135,15 @@ bool pilz::TrajectoryBlenderTransitionWindow::validateRequest(const pilz::Trajec
     return false;
   }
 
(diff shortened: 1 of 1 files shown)


jschleicher:
@PilzDE/robotics ping

martiniil (approved):

PR #172, opened 2019-06-03
drop deprecated isRobotStateEqual()

jschleicher:
* refactor existing code
* use plain C++ objects instead of creating shared pointers
* avoids interal warning facing the user for every blend request


jschleicher:
@PilzDE/robotics any comments?

martiniil (approved):

PR #173, opened 2019-06-18
Remove non-ascii char from msg/srv/action

agutenkunst:
* Cause trouble while reading via python rosbag api


martiniil (approved):
https://github.com/ros/ros_comm/issues/695#issuecomment-155239140

PR #175, opened 2019-06-21
Add test for pausing a sequence

martiniil:
* The test checks that the motion fails, since pause of sequence is not implemented yet
* Closes gap in line coverage of robot.py

### DIFF
--- a/pilz_robot_programming/src/pilz_robot_programming/robot.py
+++ b/pilz_robot_programming/src/pilz_robot_programming/robot.py
@@ -382,8 +382,6 @@ class Robot(object):
                         if isinstance(cmd, Sequence):
                             rospy.logerr("Pause not implemented for sequence yet")
                             raise RobotMoveFailed("Pause not implemented for sequence yet")
-                        # the following line is never executed due to optimization
-                        continue # pragma: no cover
                     # external stop
                     elif self._move_ctrl_sm.state == _MoveControlState.NO_REQUEST:
                         rospy.logerr("External stop of move command")
--- a/pilz_robot_programming/src/pilz_robot_programming/robot.py
+++ b/pilz_robot_programming/src/pilz_robot_programming/robot.py
@@ -382,7 +382,8 @@ class Robot(object):
                         if isinstance(cmd, Sequence):
                             rospy.logerr("Pause not implemented for sequence yet")
                             raise RobotMoveFailed("Pause not implemented for sequence yet")
-                        continue
+                        # the following line is never executed due to optimization
+                        continue # pragma: no cover
                     # external stop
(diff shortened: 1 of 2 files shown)


pilz_robot_programming/src/pilz_robot_programming/robot.py
```diff
@@ -382,7 +382,8 @@ def _move_execution_loop(self, cmd):
                         if isinstance(cmd, Sequence):
                             rospy.logerr("Pause not implemented for sequence yet")
                             raise RobotMoveFailed("Pause not implemented for sequence yet")
-                        continue
+                        # the following line is never executed due to optimization
+                        continue # pragma: no cover
```
ct2034: why is this line necessary? 
IMO the behaviour would be the same without

ct2034: I tested it. This line can be removed and doesn't change anything (Which may be the reason why it is removed by optimization ;oD)

pilz_robot_programming/test/integrationtests/tst_api_pause.py
```diff
@@ -500,6 +500,33 @@ def test_pause_and_new_command(self):
...
+        move_thread.start()
+        self.assertTrue(self.robot_motion_observer.wait_motion_start(
+            move_tolerance=self._TOLERANCE_FOR_MOTION_DETECTION_RAD, sleep_interval=self._SLEEP_TIME_S))
+
+        # 2. trigger pause
+        self.robot.pause()
+        self.assertTrue(self.robot_motion_observer.wait_motion_stop(self._WAIT_CMD_STOP_TIME_OUT_S,
+                                                                    self._TOLERANCE_FOR_MOTION_DETECTION_RAD))
+        move_thread.join()
+        self.assertTrue(move_thread.exception_thrown)
```
ct2034, line 528: I noticed that this exception is only thrown for the first time you call `pause()`. Is this the expected behaviour?

ct2034 (changes requested):

pilz_robot_programming/test/integrationtests/tst_api_pause.py
martiniil: Yes. The sequence is aborted when you try to pause it. A second `pause()` has no effect.

pilz_robot_programming/src/pilz_robot_programming/robot.py
martiniil: The effect is that execution jumps directly to the beginning of the while loop. But you are right, it is not necessary.

ct2034 (approved):

PR #187, opened 2019-07-10
Move to fixed position for brake test #186

ct2034:


ct2034:
tested it successfully on the robot with pss4000 now

martiniil (approved):
thx

PR #189, opened 2019-07-12
Add static code analyzing (clang-tidy)

hslusarek:


ct2034:
the industrial_ci supports a clang env variable ...
https://github.com/ros-industrial/industrial_ci/blob/legacy/doc/index.rst#optional-environment-variables

ct2034:
no, sorry. i think this is only for clang-format :-(
(which is != clang-tidy)

ct2034:
yes it fails :+1: 
Now we have to choose ... https://clang.llvm.org/extra/clang-tidy/checks/list.html

agutenkunst (approved):

PR #195, opened 2019-08-28
catkin and boost includes as System

agutenkunst:
Declaring catkin and boost includes as system hides them from clang tidy


ct2034:
why do we want to hide them?

ct2034 (approved):

PR #202, opened 2019-09-09
Add dependency libclang-dev to relevant packages

martiniil:


pilz_extensions/package.xml
```diff
@@ -19,6 +19,7 @@
   <url type="repository">https://github.com/PilzDE/pilz_industrial_motion</url>
 
   <build_depend>clang-tidy</build_depend>
+  <build_depend>libclang-dev</build_depend>
```
gavanderhoorn, line 22: Is this for running `clang-tidy`? Should it then not be a `test_depend`?

Same for the main `clang-tidy` dependency on the line above this one.

martiniil: clang-tidy is run during build:
https://github.com/PilzDE/pilz_industrial_motion/blob/c285668520cd8c165d7927c109c03597e4c5f354/pilz_extensions/CMakeLists.txt#L23

gavanderhoorn: Hm. That's a rather heavy dependency just for building the packages, but ok, that explains it.

agutenkunst: I opened https://github.com/PilzDE/pilz_robots/pull/239 on https://github.com/PilzDE/pilz_robots which seems more reasonable and should be sufficient to keep a clean and tidy code base.

@martiniil  @gavanderhoorn what do you think?

agutenkunst:
Closing in favor of https://github.com/PilzDE/pilz_industrial_motion/pull/203

PR #203, opened 2019-09-09
Better clang-tidy integration

agutenkunst:
* Add debs for clang-tidy to travis configuration file
* Enable clang-tidy via cmake flag

Same as https://github.com/PilzDE/pilz_robots/pull/239

I suggest using this instead of https://github.com/PilzDE/pilz_industrial_motion/pull/202 since this would require users only to have clang-tidy and libclang-dev if the want to open a PR. Also it does not require the buildfarm to install libclang-dev.

### DIFF
--- a/pilz_industrial_motion_testutils/package.xml
+++ b/pilz_industrial_motion_testutils/package.xml
@@ -16,8 +16,7 @@
   <url type="repository">https://github.com/PilzDE/pilz_industrial_motion</url>
 
   <build_depend>eigen_conversions</build_depend>
-  <build_depend>clang-tidy</build_depend>
-  <build_depend>moveit_core</build_depend>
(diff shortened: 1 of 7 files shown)


pilz_industrial_motion_testutils/package.xml
```diff
@@ -16,8 +16,7 @@
   <url type="repository">https://github.com/PilzDE/pilz_industrial_motion</url>
 
   <build_depend>eigen_conversions</build_depend>
-  <build_depend>clang-tidy</build_depend>
-  <build_depend>moveit_core</build_depend>
+    <build_depend>moveit_core</build_depend>
```
martiniil: Please fix indentation

martiniil (changes requested):

martiniil (approved):

PR #204, opened 2019-09-10
Feature/clang tidy integration

martiniil:
cherry-pick commits from melodic-devel


agutenkunst:
Seems to fail due apt being slow

agutenkunst (approved):

PR #212, opened 2019-10-01
Add client for speed_override to python API

hslusarek:
In combination with [PR #257](https://github.com/PilzDE/pilz_robots/pull/257) this PR aims to reduce the velocity for every motion (here: only in PythonAPI motions!) in such a way that the max speeds for each frame is not violated.

* Rename global_motion_factor -> speed_override
* Add client which obtains the factor via service

### DIFF
--- a/pilz_robot_programming/src/pilz_robot_programming/commands.py
+++ b/pilz_robot_programming/src/pilz_robot_programming/commands.py
@@ -208,8 +208,8 @@ class _BaseCmd(_AbstractCmd):
         # Set general info
         req.planner_id = self._planner_id
         req.group_name = self._planning_group
-        req.max_velocity_scaling_factor = self._vel_scale * robot.global_motion_factor
-        req.max_acceleration_scaling_factor = self._acc_scale * self._calc_acc_scale(robot.global_motion_factor)
+        req.max_velocity_scaling_factor = self._vel_scale * robot.speed_override
+        req.max_acceleration_scaling_factor = self._acc_scale * self._calc_acc_scale(robot.speed_override)
         req.allowed_planning_time = 1.0
 
         # Set an empty diff as start_state => the current state is used by the planner
@@ -607,8 +607,8 @@ class Gripper(_BaseCmd):
         # Set general info
         req.planner_id = "PTP"
         req.group_name = self._planning_group
-        req.max_velocity_scaling_factor = self._vel_scale * robot.global_motion_factor
-        req.max_acceleration_scaling_factor = self._acc_scale * robot.global_motion_factor
+        req.max_velocity_scaling_factor = self._vel_scale * robot.speed_override
+        req.max_acceleration_scaling_factor = self._acc_scale * robot.speed_override
         req.allowed_planning_time = 1.0
         # Set an empty diff as start_state => the current state is used by the planner
         req.start_state.is_diff = True
--- a/pilz_robot_programming/src/pilz_robot_programming/commands.py
+++ b/pilz_robot_programming/src/pilz_robot_programming/commands.py
@@ -607,8 +607,8 @@ class Gripper(_BaseCmd):
         # Set general info
         req.planner_id = "PTP"
         req.group_name = self._planning_group
-        req.max_velocity_scaling_factor = self._vel_scale * robot._speed_override
-        req.max_acceleration_scaling_factor = self._acc_scale * robot._speed_override
+        req.max_velocity_scaling_factor = self._vel_scale
+        req.max_acceleration_scaling_factor = self._acc_scale
         req.allowed_planning_time = 1.0
         # Set an empty diff as start_state => the current state is used by the planner
         req.start_state.is_diff = True
--- a/pilz_robot_programming/src/pilz_robot_programming/commands.py
+++ b/pilz_robot_programming/src/pilz_robot_programming/commands.py
@@ -208,8 +208,8 @@ class _BaseCmd(_AbstractCmd):
         # Set general info
         req.planner_id = self._planner_id
         req.group_name = self._planning_group
-        req.max_velocity_scaling_factor = self._vel_scale * robot.speed_override
-        req.max_acceleration_scaling_factor = self._acc_scale * self._calc_acc_scale(robot.speed_override)
+        req.max_velocity_scaling_factor = self._vel_scale * robot._speed_override
+        req.max_acceleration_scaling_factor = self._acc_scale * self._calc_acc_scale(robot._speed_override)
         req.allowed_planning_time = 1.0
 
         # Set an empty diff as start_state => the current state is used by the planner
@@ -607,8 +607,8 @@ class Gripper(_BaseCmd):
         # Set general info
         req.planner_id = "PTP"
         req.group_name = self._planning_group
-        req.max_velocity_scaling_factor = self._vel_scale * robot.speed_override
-        req.max_acceleration_scaling_factor = self._acc_scale * robot.speed_override
+        req.max_velocity_scaling_factor = self._vel_scale * robot._speed_override
+        req.max_acceleration_scaling_factor = self._acc_scale * robot._speed_override
         req.allowed_planning_time = 1.0
         # Set an empty diff as start_state => the current state is used by the planner
         req.start_state.is_diff = True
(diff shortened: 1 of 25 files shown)


jschleicher:
Please prefer publishers over service calls.

Generally you can advertise any value on a topic. Once a subscriber appears, it gets the regularly published value. As long as there isn't a subscriber, you have no performance overhead. And if your client is interested in a value, it probably also wants updates as continuous subscription (not only a single value from a service call).

hslusarek:
> Please prefer publishers over service calls.
> 
> Generally you can advertise any value on a topic. Once a subscriber appears, it gets the regularly published value. As long as there isn't a subscriber, you have no performance overhead. And if your client is interested in a value, it probably also wants updates as continuous subscription (not only a single value from a service call).

First of thanks for your considerations. Allow me to explain our reasoning:

- Generally we follow the recommendation in the [wiki](https://wiki.ros.org/ROS/Patterns/Communication).
- Furthermore, we try to create a logical and consistent architecture. To see our current architectural considerations, please have a look [here](https://github.com/PilzDE/pilz_robots/blob/feature/global_speed_ov/prbt_hardware_support/doc/diag_comp_overall_architecture.png).
- For the feature we try to develop here right now, it is not necessary to constantly obtain an update of the speed override and locally store the latest speed override. If needed the PythonAPI requests the latest speed override via service. The service guarantees that we obtain the latest speed override while at the same time avoiding unnecessary threading overhead in the PythonAPI itself. 
- However, please be also aware of the fact that we have an iterative development process. In other words, we mainly focus our attention on current and near feature requirements. This also means that we refactor our code, if the requirements change or new requirements emerge.

jschleicher:
From a user's point of view, I'd prefer a topic `speed_override` from `operation_mode_setup_executor -> python`. That would allow me, to subscribe to the value from a display UI node for visualization of the current scale.

martiniil:
Should we increase the API version?
https://github.com/PilzDE/pilz_industrial_motion/issues/102#issuecomment-467816895
https://github.com/PilzDE/pilz_industrial_motion/issues/102#issuecomment-499451636

jschleicher:
> 
> 
> Should we increase the API version?
> [#102 (comment)](https://github.com/PilzDE/pilz_industrial_motion/issues/102#issuecomment-467816895)
> [#102 (comment)](https://github.com/PilzDE/pilz_industrial_motion/issues/102#issuecomment-499451636)

Only breaking changes have to increase the version number. Feature additions don't.

martiniil:
> 
> 
> > Should we increase the API version?
> > [#102 (comment)](https://github.com/PilzDE/pilz_industrial_motion/issues/102#issuecomment-467816895)
> > [#102 (comment)](https://github.com/PilzDE/pilz_industrial_motion/issues/102#issuecomment-499451636)
> 
> Only breaking changes have to increase the version number. Feature additions don't.

Is it that simple? The speed override changes the behaviour of the manipulator in *reduced speed*.
Also a program fails if the user did not install a compatible version of _prbt_hardware_support_.

hslusarek:
> From a user's point of view, I'd prefer a topic `speed_override` from `operation_mode_setup_executor -> python`. That would allow me, to subscribe to the value from a display UI node for visualization of the current scale.

Please believe me I understand your reasoning completely and I'm not against your suggestion. However, regarding the UI their are still to many open questions. If all these questions are addressed, any changes necessary can and will be made. Right now, based on the current requirements it is a tossup between topic and service. I suggest we close this PR as soon as it is finished and discuss the possible UI node requirements (which are not part of this request) thoroughly in a separate meeting.

pilz_robot_programming/src/pilz_robot_programming/robot.py
```diff
@@ -124,6 +126,8 @@ def __init__(self, version=None, *args, **kwargs):
         # manage the move control request
         self._move_ctrl_sm = _MoveControlStateMachine()
 
+        self._get_speed_override_srv = rospy.ServiceProxy(Robot._GET_SPEED_OVERRIDE_SRV, GetSpeedOverride)
+
```
hslusarek: There exists a function `_establish_connections()` in which all the other services and stuff create their connection and so forth. Would it not be advisable, with regard to a clean code structure and readability, to include this connection building there too?

pilz_robot_programming/src/pilz_robot_programming/robot.py
```diff
@@ -157,32 +159,14 @@ def _robot_commander(self, robot_commander):
         self.__robot_commander = robot_commander
 
     @property
-    def global_motion_factor(self):
-        """ Defines a factor applied to all commands.
+    def speed_override(self):
```
hslusarek: I suggest making this property private, because right now, it is not intended to be used or requested by the user. Following features might change this (of course), but until these feature and their requirements are clear, I suggest keeping this property for internal use only.

pilz_robot_programming/src/pilz_robot_programming/commands.py
```diff
@@ -607,8 +607,8 @@ def _cmd_to_request(self, robot):
         # Set general info
         req.planner_id = "PTP"
         req.group_name = self._planning_group
-        req.max_velocity_scaling_factor = self._vel_scale * robot.global_motion_factor
-        req.max_acceleration_scaling_factor = self._acc_scale * robot.global_motion_factor
+        req.max_velocity_scaling_factor = self._vel_scale * robot.speed_override
+        req.max_acceleration_scaling_factor = self._acc_scale * robot.speed_override
```
hslusarek: Does the global speed override also has to apply to gripper commands? Are we missing a requirement?

pilz_robot_programming/test/integrationtests/tst_robot_functions.py
```diff
@@ -20,13 +20,13 @@
...
 from pilz_robot_programming.commands import Ptp
+from tst_api_utils import setOverrideParam
 import roslib
 
 API_VERSION = "1"
 PKG = 'pilz_robot_programming'
 roslib.load_manifest(PKG)  # This line is not needed with Catkin.
 
-
 class TestGlobalMotionFactor(unittest.TestCase):
```
hslusarek: Rename?

hslusarek (commented):
Only minor changes.

pilz_robot_programming/test/integrationtests/tst_robot_functions.py
martiniil: Done

pilz_robot_programming/src/pilz_robot_programming/robot.py
martiniil: Done

pilz_robot_programming/src/pilz_robot_programming/robot.py
martiniil: Done & added wait_for_service

pilz_robot_programming/src/pilz_robot_programming/commands.py
martiniil: I removed the scaling here.

agutenkunst:
> > > Should we increase the API version?
> > > [#102 (comment)](https://github.com/PilzDE/pilz_industrial_motion/issues/102#issuecomment-467816895)
> > > [#102 (comment)](https://github.com/PilzDE/pilz_industrial_motion/issues/102#issuecomment-499451636)
> > 
> > 
> > Only breaking changes have to increase the version number. Feature additions don't.
> 
> Is it that simple? The speed override changes the behaviour of the manipulator in _reduced speed_.
> Also a program fails if the user did not install a compatible version of _prbt_hardware_support_.

I would opt for keeping the version for now. Nevertheless I acknowledge that this is an edge case.

Finally decision @martiniil ?

leunMar:
@agutenkunst I just found out, that it is not an edge case. There are motions with very low velocity, which get infeasible due to the lower acceleration limit. :(

leunMar:
@agutenkunst Let's merge this PR and make sure that we tackle the problem with the acceleration limit and the dependency on `prbt_hardware_support` before the next release.

jschleicher:
Again: Please consider using a latched topic for speedoverride: https://github.com/PilzDE/pilz_industrial_motion/pull/212#issuecomment-537030339
That's much easier extendable from the user-perspective.

leunMar (approved):

PR #219, opened 2019-10-21
Add code coverage CI job

martiniil:

PR #221, opened 2019-10-21
Give [pilz_extensions] 100% coverage

martiniil:


pilz_extensions/include/pilz_extensions/joint_limits_interface_extension.h
```diff
@@ -52,7 +52,7 @@ inline bool getJointLimits(const std::string& joint_name,
 
   // Set the existing limits
   if(! ::joint_limits_interface::getJointLimits(joint_name, nh, limits)) {
-    return false;
+    return false; //LCOV_EXCL_LINE // The case where getJointLimits returns false is covered above.
```
agutenkunst, line 55: Then why is the check needed here?

pilz_extensions/test/integrationtest_pilz_extensions_demo.py
```diff
@@ -0,0 +1,53 @@
...
+## Test the pilz_extensions demo
+#
+# This test exists mainly for full coverage
+class IntegrationtestPilzExtensionsDemo(unittest.TestCase):
+
+    ##  Test Sequence:
+    #       1. Start demo launch file and spin until shutdown signal is received
+    #
+    #   Expected Results:
+    #       1. -
```
agutenkunst: We do a test but we don't expect anything?

pilz_extensions/test/unittest_joint_limit.cpp
```diff
@@ -43,6 +46,38 @@ TEST_F(JointLimitTest, SimpleRead) {
   EXPECT_EQ(-1, joint_limits_extended.max_deceleration);
 }
 
+/**
+ * @brief Test reading a non-existing joint limit
+ */
+TEST_F(JointLimitTest, readNonExistingJointLimit) {
```
agutenkunst: Braces!

pilz_extensions/test/unittest_joint_limit.cpp
```diff
@@ -43,6 +46,38 @@ TEST_F(JointLimitTest, SimpleRead) {
   EXPECT_EQ(-1, joint_limits_extended.max_deceleration);
 }
 
+/**
+ * @brief Test reading a non-existing joint limit
```
agutenkunst: I think this can be figured from "readNonExistingJointLimit"

pilz_extensions/test/integrationtest_pilz_extensions_demo.py
```diff
@@ -0,0 +1,53 @@
...
+    def runTest(self):
+        rospack = rospkg.RosPack()
+        path = rospack.get_path(PACKAGE_NAME)
+
+        uuid = roslaunch.rlutil.get_or_generate_uuid(None, False)
+        roslaunch.configure_logging(uuid)
+
+        launch = roslaunch.parent.ROSLaunchParent(uuid, [path + "/launch/pilz_extensions_demo.launch"])
```
agutenkunst: Could you explain why in order to check
https://github.com/PilzDE/pilz_industrial_motion/blob/57a45986764acac5df83b715df9525d257ffb72a/pilz_extensions/src/pilz_extensions_demo_node.cpp#L34-L35

a whole node is needed rather than performing this within on test?

pilz_extensions/test/integrationtest_pilz_extensions_demo.py
```diff
@@ -0,0 +1,53 @@
...
+import unittest
+import roslaunch
+import rospkg
+import rospy
+
+PACKAGE_NAME = 'pilz_extensions'
+
+## Test the pilz_extensions demo
+#
+# This test exists mainly for full coverage
```
agutenkunst: I think at least you need some more explaining to to here.

agutenkunst: Maybe this is wrong?
I applied 
```
diff --git a/pilz_extensions/CMakeLists.txt b/pilz_extensions/CMakeLists.txt
index 2ae434f..5ae605e 100644
--- a/pilz_extensions/CMakeLists.txt
+++ b/pilz_extensions/CMakeLists.txt
@@ -43,11 +43,11 @@ include_directories(
  ${catkin_INCLUDE_DIRS}
 )
 
-add_executable(${PROJECT_NAME}_demo_node src/pilz_extensions_demo_node.cpp)
+# add_executable(${PROJECT_NAME}_demo_node src/pilz_extensions_demo_node.cpp)
 
-target_link_libraries(${PROJECT_NAME}_demo_node
-   ${catkin_LIBRARIES}
- )
+# target_link_libraries(${PROJECT_NAME}_demo_node
+#    ${catkin_LIBRARIES}
+#  )
 
 #############
 ## Install ##
@@ -60,11 +60,11 @@ install(DIRECTORY include/${PROJECT_NAME}/
   PATTERN ".svn" EXCLUDE
 )
 
-install(TARGETS ${PROJECT_NAME}_demo_node
-   ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-   LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-   RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
-)
+# install(TARGETS ${PROJECT_NAME}_demo_node
+#    ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+#    LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+#    RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
+# )
 
 
 #############
@@ -87,9 +87,9 @@ if(CATKIN_ENABLE_TESTING)
      ${catkin_LIBRARIES})
 
   # Integrationtest of pilz_extensions_demo_node
-  add_rostest(test/integrationtest_pilz_extensions_demo.test)
+  # add_rostest(test/integrationtest_pilz_extensions_demo.test)
 
-  roslint_python(test/integrationtest_pilz_extensions_demo.py)
+  # roslint_python(test/integrationtest_pilz_extensions_demo.py)
 
   # to run: catkin_make -DENABLE_COVERAGE_TESTING=ON package_name_coverage
   if(ENABLE_COVERAGE_TESTING)
```
and 
`catkin_make -DENABLE_COVERAGE_TESTING=ON && catkin_make -DENABLE_COVERAGE_TESTING=ON pilz_extensions_coverage` gives me 100%. However I need to run it twice, which is strange...

agutenkunst (dismissed):

pilz_extensions/include/pilz_extensions/joint_limits_interface_extension.h
leunMar: Because the behaviour of the external `getJointLimits` might change. Then, without this check, reading the joint limits could be omitted silently.

pilz_extensions/test/unittest_joint_limit.cpp
leunMar: Done

pilz_extensions/test/unittest_joint_limit.cpp
leunMar: Done

pilz_extensions/test/integrationtest_pilz_extensions_demo.py
leunMar: For coverage of `pilz_extensions_demo_node.cpp`.

pilz_extensions/test/integrationtest_pilz_extensions_demo.py
leunMar: I don't get your point. You are commenting out the demo node? For what reason?

leunMar: I reformulated the test documentation a bit.

leunMar: Let's continue discussion here. Changes will be pushed to #222

pilz_extensions/test/integrationtest_pilz_extensions_demo.py
leunMar: Yes, we just want full coverage ... see below.

jschleicher (commented):
Another option instead of starting the demo node - that is obviously never called otherwise -  from a special test would be dropping the demo node. Why do we need `pilz_extensions_demo_node.cpp`? It isn't mentioned / linked in the [API doc](http://docs.ros.org/melodic/api/pilz_extensions/html/structpilz__extensions_1_1joint__limits__interface_1_1JointLimits.html), and the package has no README, so  first-time users cannot find it to use it as template...

leunMar:
That also was my interpretation, that the demo node serves as a template. So why dropping it? I would rather add a README. In the [API doc](http://docs.ros.org/melodic/api/pilz_extensions/html/pilz__extensions__demo__node_8cpp.html) it is already included.

agutenkunst:
I agree with @jschleicher on this one. Having this node simply add to much complexity that even I (:wink:) didn't quite grasp the usage at first. It would be hard to argue what classes / modules deserve dedicated demo_nodes.

leunMar:
> I agree with @jschleicher on this one. Having this node simply add to much complexity that even I (wink) didn't quite grasp the usage at first. It would be hard to argue what classes / modules deserve dedicated demo_nodes.

As I understand it, this demo node was intended to cover the whole package...
I just removed it.

leunMar:
@agutenkunst Squash-merge and then #219?

agutenkunst:
Ok.

jschleicher (approved):
:+1: Thanks!

PR #224, opened 2019-11-08
Feature/general speed override srv

leunMar:
The speed override is a general concept needed by the Python API. See #212.

The definition in prbt_hardware_support should be removed:
https://github.com/PilzDE/pilz_robots/blob/melodic-devel/prbt_hardware_support/srv/GetSpeedOverride.srv

### DIFF
--- a/pilz_msgs/CHANGELOG.rst
+++ b/pilz_msgs/CHANGELOG.rst
@@ -2,6 +2,11 @@
 Changelog for package pilz_msgs
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+Forthcoming
+-----------
+* Add srv definition for speed override
+* Contributors: Pilz GmbH and Co. KG
+
 0.4.7 (2019-09-10)
 ------------------
 
--- a/pilz_msgs/CMakeLists.txt
+++ b/pilz_msgs/CMakeLists.txt
@@ -2,29 +2,30 @@ cmake_minimum_required(VERSION 2.8.3)
 project(pilz_msgs)
 
 find_package(catkin REQUIRED COMPONENTS
-    message_generation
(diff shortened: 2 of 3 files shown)


jschleicher:
Re-posting into the correct PR (see https://github.com/PilzDE/pilz_industrial_motion/pull/225#issuecomment-552081727 and following comments):
Please refactor PilzDE/pilz_robots#280 to advertise a latched topic and this pr #224 would get redundant.

HiddenBug:
> Re-posting into the correct PR (see [#225 (comment)](https://github.com/PilzDE/pilz_industrial_motion/pull/225#issuecomment-552081727) and following comments):
> Please refactor [PilzDE/pilz_robots#280](https://github.com/PilzDE/pilz_robots/pull/280) to advertise a latched topic and this pr #224 would get redundant.

Your user point of view might be right but please have look at our code. We always get the [newest speed override value](https://github.com/PilzDE/pilz_industrial_motion/blob/766c606d1b13d08f6f4a7731e99902011e617b2f/pilz_robot_programming/src/pilz_robot_programming/robot.py#L166). Changing to a topic would NOT decrease complexity but actually increase it. Why? Because you have to store the last received speed override value and furthermore, you need to take care of threading issues. And like I said before in this [comment](https://github.com/PilzDE/pilz_industrial_motion/pull/212#issuecomment-537023043): 
> "[...] be also aware of the fact that we have an iterative development process. In other words, we mainly focus our attention on current and near feature requirements. This also means that we refactor our code, if the requirements change or new requirements emerge. [...]"

HiddenBug (approved):

leunMar:
@jschleicher @hslusarek I don't mind the refactoring. But currently, we have other things prioritized. I will merge this now such that we can make progress.

PR #225, opened 2019-11-08
Add brake-test srv definitions to pilz_msgs

leunMar:
Note that the brake test response message no longer contains error codes, because they are specific to the underlying service implementation and the robot hardware.
I think a message-string is sufficient. In fact, the error codes were converted into a string in the Python API anyways.

These changes can be followed up with removing the srv definitions from prbt_hardware_support, once they are released.

Plan:
1. Add msg/srv definitions to `pilz_msgs`
2. release `pilz_industrial_motion`
3. refactor `pilz_robot_programming` & `prbt_hardware_support` to use the new definitions

### DIFF
--- a/pilz_msgs/srv/BrakeTest.srv
+++ b/pilz_msgs/srv/BrakeTest.srv
@@ -1,0 +1,21 @@
+#
+# Copyright (c) 2019 Pilz GmbH & Co. KG
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+---
+bool success
+
+# Contains message explaining failure of the brake test execution
+string error_msg
--- a/pilz_msgs/srv/IsBrakeTestRequired.srv
+++ b/pilz_msgs/srv/IsBrakeTestRequired.srv
@@ -1,0 +1,22 @@
+#
+# Copyright (c) 2019 Pilz GmbH & Co. KG
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+---
+int8 result
+
+int8 NOT_REQUIRED = -1
+int8 REQUIRED = 1
+int8 UNKNOWN = 0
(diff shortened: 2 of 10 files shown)


jschleicher:
Considering latched topic (`std_msgs/Float`) would be much simpler and reduce the dependency intrinsically, see https://github.com/PilzDE/pilz_industrial_motion/pull/212#issuecomment-537030339

leunMar:
> Considering latched topic (`std_msgs/Float`) would be much simpler and reduce the dependency intrinsically, see [#212 (comment)](https://github.com/PilzDE/pilz_industrial_motion/pull/212#issuecomment-537030339)

@jschleicher I can see that this is really important to you. The service is already implemented. So it is not that simple. We would need a proper refactoring.
I guess there are better places to pose a feature request? Your comments can will be overlooked as soon as these PRs are closed.

jschleicher:
@leunMar As long as #212 is open, this feature isn't "already implemented" in my point of view. Please consider changing it now: PilzDE/pilz_robots#280 could easily do the refactoring to advertise a latched topic and this pr #225 would get redundant.

leunMar:
> Considering latched topic (`std_msgs/Float`) would be much simpler and reduce the dependency intrinsically, see [#212 (comment)](https://github.com/PilzDE/pilz_industrial_motion/pull/212#issuecomment-537030339)

How is `std_msgs/Float` suitable for communicating the results of a brake-test execution?

jschleicher:
> How is `std_msgs/Float` suitable for communicating the results of a brake-test execution?

Sorry! I mixed up speed-override with brake-test. All my comments in here are obviously out-of-scope.

pilz_msgs/srv/IsBrakeTestRequired.srv
```diff
@@ -0,0 +1,22 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+---
+int8 result
+
+int8 NOT_REQUIRED = -1
```
jschleicher: You could even abstract this by creating a Messge with just the result. See for example `moveit_msg/MoveItErrorCodes.msg` with usage in different Action/Service contexts for example in `moveit_msg/MotionPlanResponse.msg`

pilz_msgs/srv/BrakeTest.srv
```diff
@@ -0,0 +1,21 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+---
+bool success
+
+# Contains message explaining failure of the brake test execution
+string error_msg
```
jschleicher, line 21: :+1: for using a descriptive string here.

jschleicher (commented):
Probably you need a synchronous release with pilz_robots, otherwise the changes in robot.py won't allow to call the 'old' services that have different md5sums...

leunMar:
@jschleicher Thanks for your input. You are right about the md5 sums. I would like to hold back the changes in pilz_robot_programming. So the release of these changes does not need to be synchronized. After that we can do the refactoring in both repos with CI. (and then still need a synchronous release ...)

pilz_msgs/srv/BrakeTest.srv
HiddenBug, line 21: I accept this for now but only under the condition that we make a separate meeting in which we discuss a guideline on how-to properly transmit errors from ROS to the PythonAPI. This is a general issue which needs to be discussed. In my opinion, it should be consistent for all messages, etc. which are exchanged between ROS and the PythonAPI.

HiddenBug (approved):
Looks ok. Just one comment for a meeting. See review.

PR #227, opened 2019-11-15
Feature/coverage ci python

leunMar:
Add python package to CI coverage jobs

on top of #219


.coverage.sh
```diff
@@ -0,0 +1,55 @@
...
+    ici_exec_in_workspace "$extend" "$ws" catkin build $pkg -v --no-deps --catkin-make-args ${pkg}_coverage
+    cd $TARGET_REPO_PATH
+
+    if [ -a $ws/build/$pkg/${pkg}_coverage.info.cleaned ]; then
+        echo "Coverage summary for $pkg ----------------------"
+        lcov --summary $ws/build/$pkg/${pkg}_coverage.info.cleaned
+        echo "---------------------------------------------------"
+
+        line_cov_percentage=$(lcov --summary $ws/build/$pkg/${pkg}_coverage.info.cleaned 2>&1 | grep -Poi "lines\.*: \K[0-9.]*")
+    else
```
HiddenBug, line 22: I admittedly have not checked in real life what you do here. But just humor me:  What happens if I have a package without C++ or Python code or more generally an empty package with empty src folder? Do I get an error or something, even though I have no code to check?

HiddenBug (changes requested):
A question.

.coverage.sh
leunMar: You will get an error if it is included here:

>       DOCKER_RUN_OPTS='-e COVERAGE_PKGS="pilz_extensions pilz_robot_programming"'

HiddenBug: Is this the behavior we want?

leunMar: Yes. If you have no code to check, you should not include the package like above.

HiddenBug (approved):
Looks good to me. I assume you have checked locally that it fails if the coverage in python is not 100%?

leunMar:
> Looks good to me. I assume you have checked locally that it fails if the coverage in python is not 100%?

Thanks, but something is still wrong. Since I re-added pilz_trajectory_generation to the package list it does no longer work...

Yes, I did check it locally.

PR #229, opened 2019-11-22
WIP: Review cmake targets for coverage invocation

martiniil:


jschleicher:
You could consider to switch to [ros-planning/moveit_ci](https://github.com/ros-planning/moveit_ci) and use their new target with codecov.io upload.

martiniil:
> You could consider to switch to [ros-planning/moveit_ci](https://github.com/ros-planning/moveit_ci) and use their new target with codecov.io upload.

Worth considering. Thanks for the hint.

PR #231, opened 2019-11-28
Review cmake test targets

martiniil:
There exists a library `${PROJECT_NAME}_test`, which is linked to all tests. Essentially this is not enough for the integrationtests and too much for the unittests.

- [x] Link all libraries that are needed for integrationtests
- [x] Replace `${PROJECT_NAME}_test` with new library `${PROJECT_NAME}_testutils` for unittests
- [x] Make indentation of tests-block consistent

Side-effect: Reduces build-time by approx. 20s

### DIFF
--- a/pilz_trajectory_generation/CMakeLists.txt
+++ b/pilz_trajectory_generation/CMakeLists.txt
@@ -214,9 +214,9 @@ if(CATKIN_ENABLE_TESTING)
     APPEND_COVERAGE_COMPILER_FLAGS()
   endif()
 
-#########################
-####Integration Tests####
-#########################
+  #########################
+  ####Integration Tests####
+  #########################
 
   set(${PROJECT_NAME}_integrationtests
     sequence_capability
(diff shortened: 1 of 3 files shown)


pilz_trajectory_generation/CMakeLists.txt
```diff
@@ -430,7 +452,9 @@ ${catkin_LIBRARIES} ${PROJECT_NAME}_test)
   )
 
   target_link_libraries(unittest_pilz_command_planner
-    ${catkin_LIBRARIES} ${PROJECT_NAME}_test)
+    ${catkin_LIBRARIES}
+    ${PROJECT_NAME}
+  )
```
HiddenBug, line 457: Unittest `pilz_command_planner` cannot be successfully build and executed without building the complete ROS-worksapce with `catkin_make` beforehand. But maybe this was not an objective, then everything is fine with me.
(Please note I did not test every test. I just randomly picked one and tested if I can execute it without building the entire ROS-workspace)

HiddenBug (commented):
Generally it looks good to me. I'm, however, not sure if an objective was to ensure that every test can be build and run without building the complete ROS-workspace beforehand. If this was not an objective, then everything is fine with me.

HiddenBug (approved):

PR #237, opened 2020-01-23
added PoseStamped and tuple goals

rfeistenauer:
Accepts more types as goal for robot api.

- accepts PoseStamped
- accepts all iterator types (except string)
- ~~accepts JointState~~

### DIFF
--- a/pilz_robot_programming/Readme.rst
+++ b/pilz_robot_programming/Readme.rst
@@ -44,7 +44,7 @@ Robot creation
 ^^^^^^^^^^^^^^
 .. code-block:: python
 
-    from pilz_robot_programming.robot import *
+    from pilz_robot_programming import *
 
 
 At first we import the robot API.
@@ -91,7 +91,7 @@ The :py:meth:`.move` function is the most important part of the robot API. With 
 function the user can execute the different robot motion commands, like shown for :py:class:`.Ptp`, :py:class:`.Lin`
 and :py:class:`.Circ`.
 
-All cartesian goals are interpreted as poses of the tool center point (TCP) link.
+By default cartesian goals are interpreted as poses of the tool center point (TCP) link.
 The transformation between the TCP link and the last robot link can be adjusted through the ``tcp_offset_xyz`` and
 ``tcp_offset_rpy`` parameters in ``prbt.xacro``.
 
@@ -145,14 +145,16 @@ Custom Reference Frame
 ^^^^^^^^^^^^^^^^^^^^^^
 .. code-block:: python
 
-    r.move(Ptp(goal=Pose(position=Point(0, 0, 0.1)), reference_frame="prbt_tcp"))
+    r.move(Ptp(goal=PoseStamped(header=Header(frame_id="prbt_tcp"),
+                                pose=Pose(position=Point(0, 0, 0.1)))))
 
 .. code-block:: python
 
     r.move(Ptp(goal=Pose(position=Point(0, -0.1, 0)), reference_frame="prbt_link_3", relative=True))
 
-All three move classes :py:class:`.Ptp`, :py:class:`.Lin` and :py:class:`.Circ` can be executed within a custom reference frame.
-In this case, the passed goal pose will be seen relative to this coordinate system instead of the default system: ``prbt_base``
+For all three move classes :py:class:`.Ptp`, :py:class:`.Lin` and :py:class:`.Circ` you can define a custom reference frame.
+Passing a PoseStamped with the Header set to any valid tf2 ``frame_id`` is supported besides an extra argument ``reference_frame``.
+The goal passed is interpreted relative to the given coordinate frame instead of the default system ``prbt_base``
 
 The custom reference frame argument (``reference_frame="target_frame"``) has to be a valid tf frame id and can be paired with the relative command.
 When paired with relative flag, the goal will be applied to the current robot pose in this custom reference frame.
--- a/pilz_robot_programming/Readme.rst
+++ b/pilz_robot_programming/Readme.rst
@@ -152,9 +152,11 @@ Custom Reference Frame
 
     r.move(Ptp(goal=Pose(position=Point(0, -0.1, 0)), reference_frame="prbt_link_3", relative=True))
 
-For all three move classes :py:class:`.Ptp`, :py:class:`.Lin` and :py:class:`.Circ` you can define a custom reference frame.
-Passing a PoseStamped with the Header set to any valid tf2 ``frame_id`` is supported besides an extra argument ``reference_frame``.
-The goal passed is interpreted relative to the given coordinate frame instead of the default system ``prbt_base``
+For all three move classes :py:class:`.Ptp`, :py:class:`.Lin` and :py:class:`.Circ` you can define a custom reference
+frame.
+Passing a PoseStamped with the Header set to any valid tf2 ``frame_id`` is supported besides an extra argument
+``reference_frame``. The goal passed is interpreted relative to the given coordinate frame instead of the default
+system ``prbt_base``.
 
 The custom reference frame argument (``reference_frame="target_frame"``) has to be a valid tf frame id and can be paired with the relative command.
 When paired with relative flag, the goal will be applied to the current robot pose in this custom reference frame.
--- a/pilz_robot_programming/src/pilz_robot_programming/commands.py
+++ b/pilz_robot_programming/src/pilz_robot_programming/commands.py
@@ -722,9 +722,6 @@ def _is_quaternion_initialized(quaternion):
 
 def _pose_relative_to_absolute(current_pose, relative_pose):
     """Add the offset relative_pose to current_pose and return an absolute goal pose"""
-    assert isinstance(current_pose, Pose)
-    assert isinstance(relative_pose, Pose)
-
     goal_pose = deepcopy(current_pose)
 
     # translation
--- a/pilz_robot_programming/src/pilz_robot_programming/commands.py
+++ b/pilz_robot_programming/src/pilz_robot_programming/commands.py
@@ -267,6 +267,8 @@ class _BaseCmd(_AbstractCmd):
         return self._constraint_by_pose()
 
     def _constraint_by_joint_state(self):
+        if self._goal.header.stamp != rospy.Time():
+            raise ValueError("Given joint_state goal expects unsupported future execution.")
         joint_names = self._goal.name
         self._goal = self._goal.position
         return self._constraint_by_joint_values(joint_names=joint_names)
--- a/pilz_robot_programming/src/pilz_robot_programming/commands.py
+++ b/pilz_robot_programming/src/pilz_robot_programming/commands.py
@@ -130,8 +130,10 @@ class _AbstractCmd(object):
 class _BaseCmd(_AbstractCmd):
     """Base class for all single commands.
 
-    :param goal: The goal of the motion, which can be given in joint (list of float, in the order of active joints in
-        the planning group) or Cartesian space (geometry_msgs/Pose).
+    :param goal: The goal of the motion, which can be given in joint (list or tuple of float, in the order of active
+        joints in the planning group) or Cartesian space. For geometry_msgs/Pose you can specify the reference frame
+        as extra parameter `reference_frame` (see below). If a PoseStamped is passed as goal, timestamp has to be zero
+        and the `frame_id` from the Header is used instead of `reference_frame`
 
     :note:
         The geometry_msgs/Pose consists of position and orientation (quaternion). When creating an instance of
@@ -219,7 +221,7 @@ class _BaseCmd(_AbstractCmd):
         if self._goal is None:
             raise NameError("Goal is not given.")
 
-        self._check_for_future_goal_expectation()
+        self._check_header_time()
 
         convertion_methods = [self._constraint_by_joint_values,
                               self._constraint_by_pose,
@@ -236,10 +238,10 @@ class _BaseCmd(_AbstractCmd):
 
         return req
 
-    def _check_for_future_goal_expectation(self):
+    def _check_header_time(self):
         try:
-            if self._goal.header.stamp != rospy.Time():
-                raise ValueError("Given goal expects unsupported future execution.")
+            if self._goal.header.stamp != rospy.Time(0, 0):
+                raise ValueError("Given goal has unsupported time for future execution.")
         except AttributeError:
             pass
 
--- a/pilz_robot_programming/src/pilz_robot_programming/commands.py
+++ b/pilz_robot_programming/src/pilz_robot_programming/commands.py
@@ -201,8 +201,8 @@ class _BaseCmd(_AbstractCmd):
     __repr__ = __str__
 
     def _cmd_to_request(self, robot):
-        self._robot = robot
         """Transforms the given command to a MotionPlanRequest."""
+        self._robot = robot
         req = MotionPlanRequest()
 
         # Set general info
@@ -241,7 +241,7 @@ class _BaseCmd(_AbstractCmd):
             else self._robot._robot_commander.get_group(self._planning_group).get_active_joints()
         joint_values = self._get_joint_pose()
         if len(joint_names) != len(joint_values):
-            raise IndexError("Given joint goal does not match the active joints " + joint_names + ".")
+            raise IndexError("Given joint goal does not match the active joints " + str(joint_names) + ".")
         goal_constraints = Constraints()
         goal_constraints.joint_constraints = [JointConstraint(joint_name=joint_name,
                                                               position=joint_value,
--- a/pilz_robot_programming/src/pilz_robot_programming/commands.py
+++ b/pilz_robot_programming/src/pilz_robot_programming/commands.py
@@ -260,6 +260,8 @@ class _BaseCmd(_AbstractCmd):
             _to_pose_constraint(goal_pose, robot_reference_frame, self._target_link))
 
     def _set_constraint_by_stamped_pose(self, req):
+        if self._goal.header.stamp != rospy.Time():
+            raise ValueError("Given stamped pose goal expects unsupported future execution.")
         self._reference_frame = self._goal.header.frame_id if self._goal.header.frame_id != "" else _DEFAULT_BASE_LINK
         self._goal = self._goal.pose
         self._set_constraint_by_pose(req)
--- a/pilz_robot_programming/src/pilz_robot_programming/commands.py
+++ b/pilz_robot_programming/src/pilz_robot_programming/commands.py
@@ -219,16 +219,14 @@ class _BaseCmd(_AbstractCmd):
         if self._goal is None:
             raise NameError("Goal is not given.")
 
-        req.goal_constraints.append(Constraints())
-
-        convertion_methods = [self._set_constraint_by_joint_values,
-                              self._set_constraint_by_pose,
-                              self._set_constraint_by_stamped_pose,
-                              self._set_constraint_by_joint_state]
+        convertion_methods = [self._constraint_by_joint_values,
+                              self._constraint_by_pose,
+                              self._constraint_by_stamped_pose,
+                              self._constraint_by_joint_state]
 
         for method in convertion_methods:
             try:
-                method(req)
+                req.goal_constraints = method()
                 break
             except (TypeError, AttributeError, AssertionError):
                 pass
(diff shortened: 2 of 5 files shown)


gavanderhoorn:
It's up to the maintainers, but `JointState` messages are supposed to be used to encode state of a system. Current state, not desired state.

This is a bit of a pedantic thing, but using it to set a goal for a motion planner seems like it would violate semantics.

rfeistenauer:
this way you can read from joint_state topic and use this as a goal when you want to go back to a recorded position without having to extract information and check if the joint names are in the right order or for example a gripper joint is appended. (attached gripper may even be a problem. Will have to check that)

gavanderhoorn:
I understand the convenience of it, but that doesn't change the fact that it's not what this message was meant for.

Again: it's up to the maintainers. I just wanted to highlight the issue.

rfeistenauer:
Do you know a better message type, that can be used as command and has both velocities and names for joints?

jschleicher:
@gavanderhoorn moveit_commander does quite the same, see [move_group.py:506](https://github.com/ros-planning/moveit/blob/2395183dccc373eda9da9cf7ee5295d601413ceb/moveit_commander/src/moveit_commander/move_group.py#L506-L514)

If I look at the `RobotState` class implementation it seems, that many functions are related to setting goals, for example `setToRandomPositions()` and `setFromIK()`. The only data type for goals only seems to be `Constraints` used inside MotionPlanRequests...

gavanderhoorn:
@jschleicher: a *tu quoque* does not make it right :)

pilz_robot_programming/src/pilz_robot_programming/commands.py
```diff
@@ -217,40 +219,56 @@ def _cmd_to_request(self, robot):
...
+    def _set_constraint_by_pose(self, req):
+        goal_pose = self._get_goal_pose()
+        robot_reference_frame = self._robot._robot_commander.get_planning_frame()
+        req.goal_constraints[0].orientation_constraints.append(
+            _to_ori_constraint(goal_pose, robot_reference_frame, self._target_link))
+        req.goal_constraints[0].position_constraints.append(
+            _to_pose_constraint(goal_pose, robot_reference_frame, self._target_link))
+
+    def _set_constraint_by_stamped_pose(self, req):
+        self._reference_frame = self._goal.header.frame_id if self._goal.header.frame_id != "" else _DEFAULT_BASE_LINK
```
jschleicher, line 272: @agutenkunst suggested an assertion that the time in the header is zero. Otherwise users could try to set a timestamp in the future and expect the robot to move at a later point in time - which we currently ignore...

pilz_robot_programming/src/pilz_robot_programming/commands.py
```diff
@@ -189,6 +189,7 @@ def __init__(self, goal=None, planning_group=_DEFAULT_PLANNING_GROUP, target_lin
         self._relative = relative
 
         self._reference_frame = reference_frame
+        self._robot = None
```
jschleicher, line 194: Imho the robot object should neither be passed around nor set as a member object of cmds. Would it be feasible, to read the required properties in `cmd._get_sequence_request()` and only pass the current robot state and speed_override (as members of the cmd or as parameters doesn't matter for me). I'd suggest to discuss this refactoring seperately as it is out-of-scope of the current PR.

jschleicher (commented):
see inline comments.

pilz_robot_programming/test/integrationtests/tst_api_cmd_conversion.py
```diff
@@ -169,6 +172,42 @@ def test_ptp_cmd_convert_pose(self):
         self._analyze_request_general(PLANNING_GROUP_NAME, "PTP", EXP_VEL_SCALE, EXP_ACC_SCALE, req)
         self._analyze_request_pose(TARGET_LINK_NAME, exp_goal_pose, req)
 
+    def test_ptp_cmd_convert_stamped_pose(self):
+        """ Check that conversion to MotionPlanRequest works correctly.
```
agutenkunst, line 173: Please state here explicitly that setting a timestamp does not work as one may expect.

pilz_robot_programming/test/integrationtests/tst_api_cmd_conversion.py
```diff
@@ -19,14 +19,17 @@
...
+from sensor_msgs.msg import JointState
+from std_msgs.msg import Header
 from tst_api_utils import setOverrideParam
 
 
 from pilz_robot_programming.robot import *
 from pilz_industrial_motion_testutils.xml_testdata_loader import *
 from pilz_robot_programming.commands import *
 
+print "pilz_industrial_motion_testutils" in RosPack().list()
```
agutenkunst: ?

pilz_robot_programming/test/integrationtests/tst_api_cmd_conversion.py
```diff
@@ -241,6 +298,10 @@ def test_ptp_cmd_convert_negative(self):
         ptp_3 = Ptp(goal=object(), vel_scale=EXP_VEL_SCALE, acc_scale=EXP_ACC_SCALE)
         self.assertRaises(NotImplementedError, ptp_3._cmd_to_request, self.robot)
 
+        # 4
+        ptp_4 = Ptp(goal="123456", vel_scale=EXP_VEL_SCALE, acc_scale=EXP_ACC_SCALE)
```
agutenkunst, line 283:
```suggestion
        ptp_4 = Ptp(goal="some_string", vel_scale=EXP_VEL_SCALE, acc_scale=EXP_ACC_SCALE)
```

rfeistenauer: clarified the intention in using this string

pilz_robot_programming/src/pilz_robot_programming/commands.py
rfeistenauer: agree, that this may be done in a more clean way. Should i remove it out of this pr completly and go back to passing it around douzend times or leave it as is?

jschleicher: ok, let's keep it as is and defer refactoring to a follow-up PR.

jschleicher (commented):
Looks good overall. See some suggestions in my commit (still two tests failing?!)
Open question: should the `_cmd_to_request` in the Circ() and Gripper() class use the new functions as well?

rfeistenauer:
I can see it being used in Circ to accept joint interim/center, but i don't really see, where you want to use it in the gripper, except changing to the catching exception structure instead of asking for instance of. Its a single value there and no conversion is happening.

Either way I don't think one of those changes should be part of this PR. (keep it small and focused)

rfeistenauer:
> this way you can read from joint_state topic and use this as a goal when you want to go back to a recorded position without having to extract information and check if the joint names are in the right order or for example a gripper joint is appended. (attached gripper may even be a problem. Will have to check that)

As soon as an Gripper is attached, JointStates can not be read and used just as they are, since the joint_states topic publishes the gripper as well, and inserting them into the move command results in an execution error. 
To handle this we would have to match the robot joints with the incoming joint state, and throw away all unknown joints, or event execute an gripper goal as well, which in turn would probably not work in simulation ... Either way this might be unexpected behavior. Maybe we have to think about dropping the joint states after all. At least as long as this question is not answered to everyones satisfaction

jschleicher:
excluding `JointState` from the current PR to get the main purpose merged.
> As soon as an Gripper is attached, JointStates can not be read and used just as they are

pilz_robot_programming/src/pilz_robot_programming/commands.py
```diff
@@ -217,40 +219,59 @@ def _cmd_to_request(self, robot):
...
-                joint_constraint = JointConstraint()
-                joint_constraint.joint_name = joint_name
-                joint_constraint.position = joint_value
-                joint_constraint.weight = 1
-                goal_constraints.joint_constraints.append(joint_constraint)
+        convertion_methods = [self._constraint_by_joint_values,
+                              self._constraint_by_pose,
+                              self._constraint_by_stamped_pose]
 
+        for method in convertion_methods:
```
agutenkunst, line 229: I like exceptions to be exceptions.

I think the proper way would be a mapping of type to function. However rather than implementing it yourself
see if https://pypi.org/project/multimethod/ would work first.

However you need to see how this works with the "iterator types"

pilz_robot_programming/src/pilz_robot_programming/commands.py
```diff
@@ -217,40 +219,59 @@ def _cmd_to_request(self, robot):
...
         else:
             raise NotImplementedError("Unknown type of goal is given.")
 
-        req.goal_constraints.append(goal_constraints)
-
         return req
 
+    def _check_for_future_goal_expectation(self):
+        try:
+            if self._goal.header.stamp != rospy.Time():
```
agutenkunst:
```suggestion
            if self._goal.header.stamp != rospy.Time(0):
```

pilz_robot_programming/src/pilz_robot_programming/commands.py
```diff
@@ -217,40 +219,59 @@ def _cmd_to_request(self, robot):
...
+                raise ValueError("Given goal expects unsupported future execution.")
+        except AttributeError:
```
agutenkunst: since you know that the attribute you are interested in is `header.stamp` please check for it with `hasattr`

pilz_robot_programming/src/pilz_robot_programming/commands.py
```diff
@@ -217,40 +219,59 @@ def _cmd_to_request(self, robot):
...
+            except (TypeError, AttributeError, AssertionError):
+                pass
         else:
             raise NotImplementedError("Unknown type of goal is given.")
 
-        req.goal_constraints.append(goal_constraints)
-
         return req
 
+    def _check_for_future_goal_expectation(self):
```
agutenkunst: Consider renaming since this is not about future. It checks for !=0

pilz_robot_programming/test/integrationtests/tst_api_cmd_conversion.py
agutenkunst: Not resolved.

pilz_robot_programming/src/pilz_robot_programming/commands.py
agutenkunst: Have a look at it I would prefer passing it around for now, since this would only introduce the relevant changes in this PR. However I agree this needs to be looked at.

pilz_robot_programming/Readme.rst
```diff
@@ -145,14 +145,16 @@ Custom Reference Frame
...
+                                pose=Pose(position=Point(0, 0, 0.1)))))
 
 .. code-block:: python
 
     r.move(Ptp(goal=Pose(position=Point(0, -0.1, 0)), reference_frame="prbt_link_3", relative=True))
 
-All three move classes :py:class:`.Ptp`, :py:class:`.Lin` and :py:class:`.Circ` can be executed within a custom reference frame.
-In this case, the passed goal pose will be seen relative to this coordinate system instead of the default system: ``prbt_base``
+For all three move classes :py:class:`.Ptp`, :py:class:`.Lin` and :py:class:`.Circ` you can define a custom reference frame.
+Passing a PoseStamped with the Header set to any valid tf2 ``frame_id`` is supported besides an extra argument ``reference_frame``.
```
agutenkunst: Please note the constraint on rospy.Time()

agutenkunst (changes requested):
Some changes requested. See comments.

I still need to try this live...

pilz_robot_programming/src/pilz_robot_programming/commands.py
rfeistenauer: fancy, I like that, but with the newer versions python 3 is required, do we really want to use a module, that no longer is supported for python 2?

pilz_robot_programming/src/pilz_robot_programming/commands.py
rfeistenauer: it would be Time(0, 0) if you want to be all explicit, which is the default case. Is inserting the default values really best practice?

pilz_robot_programming/test/integrationtests/tst_api_cmd_conversion.py
rfeistenauer: So even with the clarified explanation in the tests description you think it would be better to test for "some_string"?

pilz_robot_programming/src/pilz_robot_programming/commands.py
jschleicher: @agutenkunst That's python, not C++: See https://docs.python.org/3.4/glossary.html EAFP and duck-typing. I had to learn that different paradigm as well...

pilz_robot_programming/src/pilz_robot_programming/commands.py
jschleicher: similar to paradigm "duck-typing": Try, if it can convert - otherwise fallback to next function looks very pythonic for me...

pilz_robot_programming/Readme.rst
jschleicher: I've updated the docstring of `BaseCommand` with that information.

Why tell in the tutorial, what *not* to do? That's a Readme and not an API spec: [don't read this](https://www.psychologytoday.com/us/blog/management-rewired/201305/dont-read).

pilz_robot_programming/test/integrationtests/tst_api_cmd_conversion.py
jschleicher: @agutenkunst Why?

pilz_robot_programming/src/pilz_robot_programming/commands.py
jschleicher: not sure about best practice, but it makes the intention much clearer.

pilz_robot_programming/src/pilz_robot_programming/commands.py
agutenkunst: Yes it could be indeed that coming from C++, this duck-typing and exception throwing is quite suprising to me.
Yet in this case it seems to me that this is also an issue for extensibility and having a simple map seems more straight forward. @martiniil I would like to hear your opinion.

pilz_robot_programming/src/pilz_robot_programming/commands.py
agutenkunst: EAFP works for me here.

pilz_robot_programming/test/integrationtests/tst_api_cmd_conversion.py
agutenkunst: Imho it would state more clearly that this could be any string not just one containing numbers. But ok.

pilz_robot_programming/src/pilz_robot_programming/commands.py
```diff
@@ -217,40 +221,59 @@ def _cmd_to_request(self, robot):
...
         return req
 
+    def _check_header_time(self):
+        try:
+            if self._goal.header.stamp != rospy.Time(0, 0):
+                raise ValueError("Given goal has unsupported time for future execution.")
+        except AttributeError:
+            pass
+
+    def _constraint_by_joint_values(self, joint_names=()):
```
agutenkunst:
```suggestion
    def _joint_values_to_goal_constraints(self, joint_names=()):
```

pilz_robot_programming/src/pilz_robot_programming/commands.py
```diff
@@ -217,40 +221,59 @@ def _cmd_to_request(self, robot):
...
+        if len(joint_names) != len(joint_values):
+            raise IndexError("Given joint goal does not match the active joints " + str(joint_names) + ".")
+        goal_constraints = Constraints()
+        goal_constraints.joint_constraints = [JointConstraint(joint_name=joint_name,
+                                                              position=joint_value,
+                                                              weight=1) for
+                                              joint_name, joint_value in zip(joint_names, joint_values)]
+        return [goal_constraints]
+
+    def _constraint_by_pose(self):
```
agutenkunst:
```suggestion
    def _post_to_goal_constraints(self):
```

pilz_robot_programming/src/pilz_robot_programming/commands.py
```diff
@@ -217,40 +221,59 @@ def _cmd_to_request(self, robot):
...
+        goal_pose = self._get_goal_pose()
+        goal_constraints = Constraints()
+        robot_reference_frame = self._robot._robot_commander.get_planning_frame()
+        goal_constraints.orientation_constraints.append(
+            _to_ori_constraint(goal_pose, robot_reference_frame, self._target_link))
+        goal_constraints.position_constraints.append(
+            _to_pose_constraint(goal_pose, robot_reference_frame, self._target_link))
+        return [goal_constraints]
+
+    def _constraint_by_stamped_pose(self):
```
agutenkunst:
```suggestion
    def stamped_pose_to_goal_constraints(self):
```

pilz_robot_programming/src/pilz_robot_programming/commands.py
martiniil: I am fine with it if you can argue the following:
- Does not catch too much
- Is not much slower than "asking for permission"

jschleicher: Yep, it just catches wrong types during conversion. I've added verbose print to display the problem to the user if conversion fails for all methods.

I'd neglect performance in this case, because it's one conversion, not thousands of commands. Even if we loose [some microseconds](https://stackoverflow.com/questions/2522005/cost-of-exception-handlers-in-python), it's not worth worrying here. But if you insist, I can still timeit to get actual numbers.

martiniil: I am not worried about the exception handlers, but:
From looking at the code I deduce that `_joint_values_to_constraint`only throws after two `robot_commander`-calls. What do you think about the delay from that?

pilz_robot_programming/src/pilz_robot_programming/commands.py
martiniil: #251

pilz_robot_programming/src/pilz_robot_programming/commands.py
jschleicher: `self._robot...` will be refactored in follow-up PR for #251.

I tend to read the start state in the conversion method as soon as possible and make it a member of `BaseCmd` to drop the robot reference... But I'll propose a follow-up PR, so we can discuss it there, if you're okay with that.

martiniil: You could also avoid unnecessary computation by checking for  `hasattr(goal, 'position')` at the beginning of `_constraint_by_pose()`and change the order of the convertion methods to:
```
        convertion_methods = [self._constraint_by_pose,
                              self._constraint_by_stamped_pose,
                              self._constraint_by_joint_values]
```

jschleicher: reordered.

(Without hasattr(), see duck-typing and EAFP above)

agutenkunst: @rfeistenauer explained that one reasoning behind this method is that something convertable by any of the three methods should also be convertable. With this aspect I would accept this way.

agutenkunst (approved):
Sorry this took so long. Thank you for your contribution!

agutenkunst:
@martiniil please merge if you agree. Since you are "pending reviewer" I don't want to pass over you.

pilz_robot_programming/src/pilz_robot_programming/commands.py
martiniil: I am not convinced that duck-typing works here:
```
    def _joint_values_to_constraint(self, joint_names=()):
        if isinstance(self._goal, str):
            raise TypeError("String is not convertible into joint values.")
```

pilz_robot_programming/src/pilz_robot_programming/commands.py
```diff
@@ -217,39 +221,58 @@ def _cmd_to_request(self, robot):
         if self._goal is None:
             raise NameError("Goal is not given.")
```
martiniil, line 222: Can we get rid of these lines? (EAFP)

jschleicher: yes, of course. Just go ahead. Or put it into a follow-up PR, since it is out-of-scope / unchanged in the current PR.

martiniil: Imho it is not out-of-scope. You changed how the member `self._goal` is treated inside this method. We should keep it somehow consistent. Of course we cannot change the coding-style of the whole API...

PR #242, opened 2020-02-06
Fix CodeCoverage warnings

martiniil:
analogue to https://github.com/PilzDE/pilz_robots/commit/c9cbba049104b942c5a4c328a38d8e5f17f9a523


agutenkunst:
Seems like there is something wrong with the coverage of pilz_trajectory_generation.
Either it was wrong before but not discovered or it is calculated wrong now.

@martiniil can you have a look at it?

agutenkunst (approved):
Thank you!

PR #246, opened 2020-02-17
Print position limit error reason.

DanBmh:


ct2034:
we are currently in the process of merging the planner into moveit https://github.com/ros-planning/moveit/pull/1893. I would like to put this on hold until we are done with that. Could you please open your PR there, when we are done with the merge.

Issue #95, opened 2019-02-18
Relative movement in sequence is broken


agutenkunst:
When using relative motion in a sequence the relative seems to reference to the position before the sequence. This is especially bad when transitioning from non-sequence to sequence since the robot could crash.

### Commit
pilz-robot-programming 0.3.3

### Steps to reproduce
Run
```
#!/usr/bin/env python

from geometry_msgs.msg import Pose, Point
from pilz_robot_programming.robot import *
from pilz_robot_programming.commands import *
import math
import time

def start_program():
    print("Executing " +__file__)

    #initial variables
    r = Robot('1')
    vel = 0.1

    orientation_pick = from_euler(0, math.radians(180), math.radians(45))
    
    r.move(Ptp(goal=Pose(position=Point(-0.499, -0.195, 0.3), orientation=orientation_pick), vel_scale = vel, relative=False))

    print("High 1 before: ", r.get_current_pose().position.z)
    r.move(Lin(goal=Pose(position=Point(0,0,-0.05)), vel_scale = vel, relative=True))
    print("Low 1: ", r.get_current_pose().position.z)
    r.move(Lin(goal=Pose(position=Point(0,0,0.05)), vel_scale = vel, relative=True))
    print("High 1 after: ", r.get_current_pose().position.z)
    

    seq = Sequence()
    seq.append(Lin(goal=Pose(position=Point(0,0,-0.05)), vel_scale = vel, relative=True))
    seq.append(Lin(goal=Pose(position=Point(0,0,0.05)), vel_scale = vel, relative=True))
    print("High 2 before: ", r.get_current_pose().position.z)
    r.move(seq)
    print("High 2 after: ", r.get_current_pose().position.z)

if __name__ == "__main__":
    # Init a rosnode
    rospy.init_node('robot_program_node', anonymous=True)

    start_program()
```
### Expected behavior
The two seperate LIN should move the robot exactly in the same way like the two LIN in the sequence.

### Observed behavior
During the sequence the robot moves 0.05 down but 0.1 up.

### Logfiles
Output of `rostopic echo /sequence_move_group/goal`
[output_goal.txt](https://github.com/PilzDE/pilz_industrial_motion/files/2876024/output_goal.txt)

Output in console where program runs:
```
Executing ./row_pick_place.py
('High 1 before: ', 0.29999482722152604)
('Low 1: ', 0.25003443220807708)
('High 1 after: ', 0.30001076941836813)
[...]

PR #253, opened 2020-02-28
store only the relevant members instead of robot object

### LINKED ISSUE #251: Remove member robot_ from commands
See: https://github.com/PilzDE/pilz_industrial_motion/pull/237#discussion_r374143742

Robot object should neither be passed around nor set as a member object of cmds.

jschleicher:
to avoid using robot functions, when we only want to access some data

Fixes #251 

Reading the arguments could be pulled out one more layer into `_get_sequence_request`, but that would have to change all tests in tst_api_cmd_conversion.py - so I wait for your feedback on this step first.


martiniil (commented):
@jschleicher What is the rational behind not passing the robot object to the convertion methods?

jschleicher:
@martiniil [Loose coupling](https://www.tutorialspoint.com/what-are-the-differences-between-tight-coupling-and-loose-coupling-in-java): Why should commands be able to use the whole robot api, when they only need to know (and transform) the start state?

Note that this refactoring is a first step towards fixing #95 as well, because the last goal could be easily set as next start in a sequence.

martiniil:
> @martiniil [Loose coupling](https://www.tutorialspoint.com/what-are-the-differences-between-tight-coupling-and-loose-coupling-in-java): Why should commands be able to use the whole robot api, when they only need to know (and transform) the start state?
> 
> Note that this refactoring is a first step towards fixing #95 as well, because the last goal could be easily set as next start in a sequence.

Thanks. I approve this idea. Concerning the code I have only one comment.

pilz_robot_programming/src/pilz_robot_programming/commands.py
```diff
@@ -204,14 +204,19 @@ def __str__(self):
...
     def _cmd_to_request(self, robot):
         """Transforms the given command to a MotionPlanRequest."""
-        self._robot = robot
         req = MotionPlanRequest()
 
+        self._robot_reference_frame = robot._robot_commander.get_planning_frame()
+        self._active_joints = robot._robot_commander.get_group(self._planning_group).get_active_joints()
+        self._start_joint_states = robot.get_current_joint_states(planning_group=self._planning_group)
+        self._start_pose = robot.get_current_pose(target_link=self._target_link, base=self._reference_frame)
+        self._tf_listener = robot.tf_listener_
```
martiniil: How about adding methods `get_reference_frame()` and `get_active_joints()` to the Robot class?
or:
Inject a method for the pose transformation by the robot? This would loose the coupling even more.

martiniil (changes requested):

pilz_robot_programming/src/pilz_robot_programming/commands.py
jschleicher: done.

jschleicher:
@martiniil What do you think about reading the properties up inside `_get_sequence_request` already? Or would you approve as-is?

martiniil:
> @martiniil What do you think about reading the properties up inside `_get_sequence_request` already? Or would you approve as-is?

This would lead to code duplication in the BaseCmd/Sequence class. Let's keep it like that.

martiniil (approved):

PR #255, opened 2020-03-27
Attribute based equivalence for commands

rfeistenauer:
I have implemented a command equivalence, that does not just compare the object, but its attributes. 
e.g currently: 
```
Ptp() == Ptp() 
False
```
with this PR its True:
```
Ptp() == Ptp()
True
Ptp() == Lin()
False
Ptp(goal=[1, 2]) == Ptp(goal=[1, 2])
True
```

Needed this to greatly reduce the effort for comparison in some tests.
Are there any arguments against including this as a feature into the api?

### DIFF
--- a/pilz_robot_programming/src/pilz_robot_programming/commands.py
+++ b/pilz_robot_programming/src/pilz_robot_programming/commands.py
@@ -185,7 +185,13 @@ class BaseCmd(_AbstractCmd):
         # Needs to be set by derived classes
         self._planner_id = None
 
-        self._goal = goal
+        try:
+            if isinstance(goal, str):
+                raise TypeError()
+            self._goal = tuple(goal)
+        except TypeError:
+            self._goal = goal
+
         self._planning_group = planning_group
         self._target_link = target_link
         self._vel_scale = vel_scale
@@ -203,15 +209,7 @@ class BaseCmd(_AbstractCmd):
 
     def __eq__(self, other):
         if isinstance(other, BaseCmd):
-            return self._goal == other._goal \
-                   and self._planner_id == other._planner_id \
-                   and self._planning_group == other._planning_group \
-                   and self._target_link == other._target_link \
-                   and self._vel_scale == other._vel_scale \
-                   and self._acc_scale == other._acc_scale \
-                   and self._relative == other._relative \
-                   and self._reference_frame == other._reference_frame
-
+            return hash(self) == hash(other)
         return NotImplemented
 
     def __ne__(self, other):
@@ -221,7 +219,7 @@ class BaseCmd(_AbstractCmd):
         return NotImplemented
 
     def __hash__(self):
-        return hash((hash(t) for t in tuple(sorted(self.__dict__.items()))))
+        return hash(tuple(sorted([str(t) for t in self.__dict__.items()])))
 
     __repr__ = __str__
 
@@ -390,9 +388,9 @@ class Ptp(BaseCmd):
         out_str = BaseCmd.__str__(self)
         if self._relative:
             out_str += " relative: True"
-        if isinstance(self._goal, Pose):
+        if isinstance(self._goal, Pose) or isinstance(self._goal, PoseStamped):
             out_str += " Cartesian goal:\n" + str(self._goal)
-        if isinstance(self._goal, list):
+        elif isinstance(self._goal, tuple):
             out_str += " joint goal: " + str(self._goal)
         return out_str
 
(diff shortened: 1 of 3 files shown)


pilz_robot_programming/src/pilz_robot_programming/commands.py
```diff
@@ -478,6 +500,14 @@ def __init__(self, interim=None, center=None, vel_scale=_DEFAULT_CARTESIAN_VEL_S
         self._interim = interim
         self._center = center
 
+    def __eq__(self, other):
+        if isinstance(other, Circ):
+            return super(Circ, self).__eq__(other) \
+                   and self._interim == other._interim \
+                   and self._center == other._center
+
+        return NotImplemented
```
ct2034:
```suggestion
        else:
            return NotImplemented
```
makes it more readable

pilz_robot_programming/src/pilz_robot_programming/commands.py
```diff
@@ -201,6 +201,28 @@ def __str__(self):
...
+
+    def __ne__(self, other):
+        x = self.__eq__(other)
+        if x is not NotImplemented:
+            return not x
+        return NotImplemented
+
+    def __hash__(self):
+        return hash(tuple(sorted(self.__dict__.items())))
```
ct2034:
```suggestion
        return hash(tuple(map(lambda k: self.__dict__[k], sorted(self.__dict__.keys()))))
```
I did not test it, but this should make https://github.com/PilzDE/pilz_industrial_motion/pull/255/files#diff-288808388fd41c35fcc136672ed08a47R207-R213 not necessary

ct2034 (changes requested):
These changes are super cool and definitely helpful. I have two suggestions in the code.
Generally, please also add unit tests, then we could see how my suggestions perform ;-)

rfeistenauer:
Thank you for your reply. I waited with the tests until someone approved this change. Will provide them shortly in this case. Ill also look into your changes then.

pilz_robot_programming/src/pilz_robot_programming/commands.py
rfeistenauer: I do not really agree, that adding extra stuff to read makes it more readable. Id rather just delete the empty line.

pilz_robot_programming/src/pilz_robot_programming/commands.py
rfeistenauer: I had to change the method. Did try to change and remove/replace code like you suggested anyway, but I didn't find a working solution. Either provide more precise input, of what to delete or replace, or test it yourself with the provided tests please.

pilz_robot_programming/src/pilz_robot_programming/commands.py
rfeistenauer: Is this with the new code still an issue?

pilz_robot_programming/src/pilz_robot_programming/commands.py
```diff
@@ -185,7 +185,13 @@ def __init__(self, goal=None, planning_group=_DEFAULT_PLANNING_GROUP, target_lin
         # Needs to be set by derived classes
         self._planner_id = None
 
-        self._goal = goal
+        try:
+            if isinstance(goal, str):
+                raise TypeError()
```
rfeistenauer, line 190: Should probably throw an `NotImplementedError` in the furture, but didn't want to introduce an api break

pilz_robot_programming/src/pilz_robot_programming/commands.py
```diff
@@ -417,9 +437,9 @@ def __str__(self):
         out_str = BaseCmd.__str__(self)
         if self._relative:
             out_str += " relative: True"
-        if isinstance(self._goal, Pose):
+        if isinstance(self._goal, Pose) or isinstance(self._goal, PoseStamped):
             out_str += " Cartesian goal:\n" + str(self._goal)
-        if isinstance(self._goal, list):
+        elif isinstance(self._goal, tuple):
```
rfeistenauer, line 442: This is already not sweet, and will get quite messy, with more types. We should try to find an alternative.

pilz_robot_programming/src/pilz_robot_programming/commands.py
rfeistenauer: more or less implemented your change, you can check if you are OK with my implementation.

jschleicher (approved):
LGTM

jschleicher:
@ct2034 Please Re-review or dismiss your old review.

ct2034 (approved):

PR #256, opened 2020-04-01
makes single instance lock claim and releasable

rfeistenauer:
This is a feature proposal, that would change the current single instance behavior to one, that allows to keep several instances, while only one is allowed to trigger move requests. This allows a clearer separation of responsibilities spread across ROS nodes.

The Robot still acts, as it did up to now, with the difference, that the node keeping the main instance is able to release and reclaim an instance without having to recreate the whole Instance, which has quite some overhead. In the meantime another node can create and hold an instance without the right to move it, thus is able to read the robot state for example or wait for the other node to finish before starting its own process.

Please state what you think about this feature.

### DIFF
--- a/pilz_robot_programming/src/pilz_robot_programming/robot.py
+++ b/pilz_robot_programming/src/pilz_robot_programming/robot.py
@@ -114,7 +114,7 @@ class Robot(object):
     # time constant
     _SERVICE_WAIT_TIMEOUT_S = 1
 
-    def __init__(self, version=None, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


rfeistenauer:
As alternative or additional possibility to trigger events from different nodes I could provide a service wrapper. I used that in a first try, but replaced it with the solution provided in this PR, since its much simpler to use and maintain.

ct2034:
I would suggest putting this on hold until we work on the *single-point-of-control*. This is one task that will come up in the ISO10218 context and this one would be a good aspect in the discussion about the technical realization we have to have then.

PR #282, opened 2020-05-14
Add storage feature

### LINKED ISSUE #301: Merge PR#282 into melodic-devel
Is waiting for #291

hslusarek:
PR PilzDe/pilz_teach#59 removes storage feature from pilz_teach repo.


pilz_store_positions/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+
+#############
+## Install ##
+#############
+
+catkin_python_setup()
+
+catkin_install_python(PROGRAMS
+   scripts/pose_visualisation_node.py
+   scripts/store_current_pose.py
```
jschleicher: While you're at renaming: Maybe consider to drop the `.py` suffix as suggested in http://docs.ros.org/melodic/api/catkin/html/howto/format2/installing_python.html, if these are user-executable scripts?

pilz_store_positions/src/pilz_store_positions/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from .ros_message_serializer import RosMessageSerializer
+from .pose_file_tf_publisher import PoseFileTFPublisher
```
jschleicher, line 17: Maybe `import *` as in [pilz_robot_programming](https://github.com/PilzDE/pilz_industrial_motion/blob/kinetic-devel/pilz_robot_programming/src/pilz_robot_programming/__init__.py)?  or [tf2_ros](https://github.com/ros/geometry2/blob/noetic-devel/tf2_py/src/tf2_py/__init__.py)

jschleicher (commented):
see inline comments

hslusarek:
@rfeistenauer Is the dependency to our PythonAPI really necessary, see line 31 in [`store_current_pose.py`](https://github.com/PilzDE/pilz_teach/blob/66e0aeb41b8f8b5e33d1cb04009b503a2d77fef2/pilz_teleoperation/scripts/store_current_pose.py#L31)? Maybe it is better to separate the `get_current_pose()` function into a separate python script. Using our PythonAPI requires that our planner is used because, otherwise, the script won't run.

jschleicher:
> Is the dependency to our PythonAPI really necessary

You're absolutely correct. This could be extracted in the future. I'd suggest to keep things simple in the current PR and follow-up with refactoring in a next step.

pilz_store_positions/src/pilz_store_positions/__init__.py
ct2034: If there are multiple classes to be imported, `*` makes sense, yes. In this case we only need one class per file so I think is it is cleaner to import them named.

rfeistenauer:
Should be possible to use tf listener instead or improve / split that info out of robot class

agutenkunst (approved):
This has a few pragmatic solutions in it in order to get things done. But PRs https://github.com/mikeferguson/code_coverage/pull/23 and https://github.com/machinekoder/ros_pytest/pull/6 are in progress for a proper and cleaner solution. Thanks to everybody contributing in this. Approved.

agutenkunst:
I re-requested a review from @jschleicher . If @jschleicher is fine with it please anybody squash + merge.

jschleicher (approved):
Approving.

PR #303, opened 2020-06-02
Use named jobs in travis

agutenkunst:
Easier to read in the overview.

**Before** No indication of what the job is about:
Screenshot from 2020-06-02 18-00-56

**After** much better:
Screenshot from 2020-06-02 17-59-40


hslusarek (approved):
Looks good to me.

PR #317, opened 2020-10-01
update package.xml for python3

jschleicher:
a tiny step towards noetic compatibility

Note: This needs https://github.com/ros/rosdistro/pull/26778 as prerequisite.


martiniil:
Thanks. We made use of this in #331 (except for the format version). Since we don't need these changes for melodic: Close?

jschleicher:
@martiniil Why not use the `condition` attribute to keep the difference small between melodic and noetic branches?

